PR #1, opened 2019-12-04
Remove GetSvnInfo()

aismann:


tajmone:
I think that historically PureBasic has always been developed using Subversion — at least, I've noticed SVN related files in the Linux distro of PB. 

So, maybe that procedure is there for the sake of the PB Team workflow.

alphasnd:
Yes, I think it won't be needed anymore as we will handle branches differently from now on (always patch master and create a branch when releasing a new PB version). Thanks for the patch !

PR #2, opened 2019-12-06
Add Git Configurations

tajmone:
This commit adds `.gitattributes` and `.gitignore` settings optimized for PureBasic projects in general, and for the current project specifically.

Adapted from the "PureBasic Repository Template" for GitHub (cc0):

* https://github.com/tajmone/purebasic-repository-template

These settings will ensure cross platform conflicts due to end-of-line differences by enforcing Git to check-in all files with `LF` EOL and handle EOL normalization at check-out according to the settings and/or the native EOL of end uesers' OS.


alphasnd:
Thanks !

fantaisie-software:
I just merged it, should we see color highlightning in GitHub for .pb files ? It doesn't seems to show here

tajmone:
> I just merged it, should we see color highlightning in GitHub for .pb files ? It doesn't seems to show here

No, currently there is no PB syntax available for GitHub, but at least it doesn't get wrongly highlighted via the syntax guesser using another language. So, currently it serves mainly for repo stats and preventing bad highlighting.

If in the future GitHub were to support PB highlighting, the repo will work straight away thanks to the extra Linguist settings.

I've created various PureBasic syntaxes in the years:

- [`purebasic.lang`][purebasic.lang] for [Highlight]
- [`purebasic.js`][purebasic.js] for [highlightjs]
- [`PureBasic.sublime-syntax`][PureBasic.sublime-syntax] (WIP draft) for [Sublime Text]

So I had thought of creating a syntax definition that would work on GitHub — if I remember correctly, GitHub relies on [Rouge] for syntax highlighting, but  there are also various options beside Rouge, including TextMate syntaxes.

The problem is maintaining the syntax up to date, for the full list of PureBasic keywords and tokens is not available anywhere (see my _[Need The Full List of PureBASIC Keywords]_ post on PB Forum on this topic).

So far, the only way I've managed to update the PB syntaxes I've created is by parsing the Highlighter DLL that ships with the PB Windows SDK — admittedly, it's a rather cumbersome and inelegant solution, but currently there is no other viable way to obtain the full list of keywords for each PB version.

When it comes to Editors, one needs to target only the syntax tokens of the current PB release, but with highlighters you need to cover all tokens from every release, because highlighter might be used for old and new code alike.

Hopefully now that the PB IDE is open source I can finally get access to the full set of PB keywords directly from its sources, instead of having to hack my way through the compiled DLL. This would make my life much easier, for due to the cumbersome process of DLL hacking I haven't always found the time to update the various syntaxes as soon as a new PB release was out (indeed, some syntaxes are lagging behind). If I could automate the whole process it would be much better.

Also, I might be able to create a SpiderBasic syntax too, for the same problem applies to SB, except that I couldn't find an highlighter DLL to extract the keywords from.

I don't even know if anyone has ever used any of those syntaxes, for I never got any feedback in all these years, but since I use them for my own projects I'd like to keep them up to date. Besides, being the co-maintainer of the [Highlight repository], I'd like to keep all syntaxes up to date wherever possible.

[purebasic.lang]: https://gitlab.com/saalen/highlight/blob/master/langDefs/purebasic.lang
[purebasic.js]: https://github.com/highlightjs/highlight.js/blob/master/src/languages/purebasic.js
[PureBasic.sublime-syntax]: https://github.com/tajmone/sublime-purebasic/blob/master/PureBasic.sublime-syntax

[Highlight]: http://www.andre-simon.de/
[highlightjs]: https://highlightjs.org/
[Sublime Text]: https://github.com/tajmone/sublime-purebasic

[Rouge]: https://github.com/rouge-ruby/rouge

[Need The Full List of PureBASIC Keywords]: https://www.purebasic.fr/english/viewtopic.php?f=13&t=68368
[Highlight repository]: https://gitlab.com/saalen/highlight/

PR #4, opened 2019-12-07
Editor config setup

tajmone:
These two commits introduce [EditorConfig] support to help enforcing code styles consistency in the repository across various editors and IDEs.

Most modern editors natively support EditorConfig, and free plug-ins are available for those who don't. End users don't have to do anything, for the editor will automatically pick-up the settings in `.editorconfig`, overriding the user settings in favour of the project specific settings.

In its current state, the repository mixes indentation styles (tabs vs spaces) in the same files, and use of EOLs and indentation width varies across same file types and even within the same files. Many files contained trailing whitespace.

Adoption of EditorConfig will enforce code styles consistency and best practices across the repository, on per-file type basis, with the possibility of adopting folder-specific settings if the need be due to third party code inclusion in the project.

I've manually cleaned up all the existing files so they'd pass EditorConfig validation via [EClint].

I've also set-up the repository to use [Travis CI] to test that every commit and PR meets the code styles set by EditorConfig, via [EClint]. This should be a time-saver for maintainers because badly formatted code will be immediately detectable via the Travis Status report icon.

The build system is working fine on my fork (which is Travis-enabled):

- https://github.com/tajmone/purebasic/runs/338108266

I hope that these commits can be the first steps toward improving the collaborative aspect of the repository. Lack of code styles consistency tends to make diffing harder to read, create spurious commits due to Git picking up whitespace contents changes, and ultimately it makes batch fixes to the code base much harder to accomplish via RegEx and tools like SED.

Currently, I see potential problems with the PB sources when it comes to collaborative editing. I'm not quite sure which editor was used for this project, but the lack of a BOM in PB sources is going to be a problem when users start editing files with the PureBasic IDE (which enforces the BOM on UTF-8). 

Another problem I can foresee is with arbitrary indentation found in the current sources — which is likely to be flatted-out when users re-format the code with PB IDE. Being able to lint code using a standard tool is a good way to preserve code consistency, and the PB IDE should be the reference tool for reformatting PB sources, since every PB user has access to it.

Anyhow, I'll create some dedicated issues to address in detail collaborative editing problems and various pitfalls relating to the use of PB with Git (and the PB IDE settings that can mitigate them). We definitely need to add some contributors guidelines to the project.

[EditorConfig]: https://editorconfig.org/
[EClint]: https://www.npmjs.com/package/eclint
[Travis CI]: https://travis-ci.com/


fantaisie-software:
May be the best would be to add a BOM to every .pb/pbi file and use bom-utf8 in EditorConfig ? We wrote the IDE and tool with the IDE didn't support UTF8 so it remained like that, but now it should use BOM or we can't use the PB IDE to edit files. Could you amend your commit to add BOM ?

tajmone:
OK, I've added the BOM and tweaked the EditorConfig settings for PB files encodings to `charset = utf-8-bom`.

Now there only remains to decide if we ought to reformat all files using the PB IDE linter or not.

I also propose than once all issues are dealt with we squash the last two commits befor merging into `master` (keeping the EditorConfig settings and source cleanup as separate commits might be a good idea, just in case someone needs to cherry pick the EditorConfig part into their own forks to catch up with upstream changes).

fantaisie-software:
OK for me to use the PB Indent to format the code, feel free to amend and force push :)

tajmone:
Done! 

In the end, I've only amended the last commit, so we can keep the three commits
separate, in case someone need to rely on any one of them:

- 3b40c01b — EditorConfig Setup
- 8e133f49 — Fix Code Styles
- 34f9eced — Normalize PB Sources

Thos who rushed to work on the IDE sources might need one or more of these commits to rebase their changes before submitting a contribution, so it might be best to keep them as individual commits.

I think we're ready merge now.

PR #29, opened 2019-12-15
Add Images Previews in README Files

tajmone:
These commits all relate to adding README files to various subfolders, providing info about their contents and image previews.

The goal is to improve the navigation experience and contents discovery on GitHub WebUI, reducing the overwhelming impact of a project that contains so many files. The ultimate goal is to provide in each folder a README file that allows end users to grasp what's there and what does what, without having to leave the page in order to peek at previews of single files (code or image) or folders.

I've kept the commits separate, instead of squashing them together into a huge single commit, because this is a long term task and more commits of this type are on the way anyhow, so it's worth keeping them to human-manageable sizes.

### DIFF
--- a/PureBasicDebugger/Data/README.md
+++ b/PureBasicDebugger/Data/README.md
@@ -1,0 +1,71 @@
+# Debugger GUI Images
+
+The icon set for the [PureBasic debugging tools] GUI elements.
+
+-----
+
+**Table of Contents**
+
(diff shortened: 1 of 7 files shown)


tajmone:
Hopefully all these `README.md` files, with descriptions and images previews, should make it easier to navigate the repository and get a feeling of what its contents are.

PR #33, opened 2019-12-18
Fix the weekly update check

SicroAtGit:
Forum thread: [PB-IDE: No automatic update notification](https://www.purebasic.fr/english/viewtopic.php?f=23&t=73609)

Please note that the fix was not tested.


fantaisie-software:
Thanks for the fix !

PR #34, opened 2019-12-20
Purge empty sessions from Session History list

kenmo-pb:
Purge sessions from the Session History list if they don't have any contents.
These "empty" sessions just clutter up the list. Currently you need an IDE tool to remove them.
https://www.purebasic.fr/english/viewtopic.php?f=3&t=65019
https://www.purebasic.fr/english/viewtopic.php?f=3&t=73336

In this pull request, this is enabled or disabled at compile-time by a new constant.

If there is really some value in showing "empty" sessions, then I can extend this feature to be a user-toggled setting. Thoughts?

### DIFF
--- a/PureBasicIDE/EditHistory.pb
+++ b/PureBasicIDE/EditHistory.pb
@@ -16,6 +16,9 @@ UseSQLiteDatabase()
 ; disable this only for testing
 #HISTORY_WRITE_ASYNC = #True
 
+; enable to purge sessions which don't contain any edit events
+#HISTORY_PURGE_EMPTY_SESSIONS = #True
+
 ; keep previous versions of displayed sources cached
 ; to speed up loading (as reconstructing multiple diffs takes time)
 ; especially when browsing within the history of one file, this helps a lot
@@ -441,7 +444,6 @@ CompilerIf #CompileWindows
       ; Now purge old sessions if needed
       ;
       PurgeList$ = ""
-      FirstPurged = #True
       
       If HistoryPurgeMode = 1
         
@@ -460,12 +462,7 @@ CompilerIf #CompileWindows
           
           ; record sessions for deleting
           While NextDatabaseRow(#DB_History)
-            If FirstPurged
-              FirstPurged = #False
-            Else
-              PurgeList$ + ", "
-            EndIf
-            PurgeList$ + Str(GetDatabaseLong(#DB_History, 0))
+            PurgeList$ + ", " + Str(GetDatabaseLong(#DB_History, 0))
           Wend
           
           FinishDatabaseQuery(#DB_History)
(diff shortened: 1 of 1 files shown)


tajmone:
> Thoughts?

It would be useful to be able to quickly toggle on/off this feature for the current session. I often work on the same PB sources with different editors beside the PB IDE, e.g. Sublime Text which offer multi-cursor editing and tons of other functionalities. Each time I save a file in Sublime Text, when the PB IDE gains focus again it warns about files having been changed and asks refresh confirmation.

The refresh part is fine (although it would be usefeul to be able to also quickly toggle this between no-confirmation and ask for confirmation), but all these spurious "snapshots" are really overkill for the session history.

In general, when switching to another editor it would be nice to have a quick way to temporarily disable a certain group of functionalities, or alter their default behavior, and then switch back to normal mode. Right now I'm thinking out of my head, without the PB IDE open in front of me, so I might be missing out and misrepresenting some details, but the overall concept should be clear.

This might also apply to version control, e.g. when switching between branches — you wouldn't want to mix in the same session history files from different development branches. So there are a lot of implications at stake here if considering using the PB IDE in a wider context than its normal use (but one might argue correctly that these are edge-cases, not falling within the general scope of the IDE).

kenmo-pb:
I think you're combining two different features.

I'm talking about the `File` > `Session History` dropdown list of past IDE sessions. The dropdown shows sessions even if they have no contents, meaning you opened and closed the IDE without editing any files. In my opinion, that just clutters up the list.

You're talking about editing files outside of the IDE. There's already a setting to turn off the warning (`Preferences` > `Editor` > `Monitor open files for changes on disk`) but I agree, there should be a third option `Reload automatically` or `Reload silently` or `Reload without confirmation` etc. Notepad++ calls it `Update silently`.

An alternative to this Pull Request I submitted would be: keep empty sessions in the database, but give the user a toggle of whether to show or hide them. Maybe that's better?

tajmone:
> An alternative to this Pull Request I submitted would be: keep empty sessions in the database, but give the user a toggle of whether to show or hide them. Maybe that's better?

I'm not sure whether there are any benefits of storing empty session at all. Are there?

fantaisie-software:
Yes I think it would be better to not record empty sessions at all..

kenmo-pb:
Yes I see three options:

1. Purge empty sessions from database (as I submitted PR)
2. Keep empty sessions in database — but hide them from list, or toggle them
3. Never add empty sessions to the database

The *only reason* I can think for keeping empty sessions is: maybe tracking crashed IDE sessions? But that seems like a silly usage of the Session History.

Option (1) is nice because there was already a "purge" section in the code, I just added to it.
Option (3) is good too, but what if the IDE crashes, or killed by OS shutdown, etc.? Will you find empty session entries "left over"? I need to look at that, and test.

fantaisie-software:
I think point 1. and 3. should be done to solve this issue

kenmo-pb:
Good news!

After checking the code, and testing many cases, I can confirm that this pull request *already achieves Point 3*, in addition to Point 1 (the original goal).

Since the "purge empty sessions" happens after the current session is "closed" in the database, the current session will immediately be purged if it has no edit events.

So I believe this is OK to merge.

fantaisie-software:
Thanks !

Issue #15, opened 2019-12-09
Safeguards Against PureBasic IDE Settings


tajmone:
We need to address some potential interferences with Git by the PureBasic IDE changing source files at save time, which could lead to false-positive modified files in Git working directory.

The main sources of this problem are:

1. A contributor's IDE being configured to "__Save Setting to__ » __The end of the Source file__".
2. The presence in the repository of PB project files (`.pbp`) being tracked.

# Save Setting to

The first problem would cause PB sources to be constantly modified by each end-user that opens the file in the PB IDE, for the IDE would store user-associated info (like PBCompiler version, cursor position, etc.) at the end of the source file, causing Git to detect it as modified even though the user didn't actually edit its contents (I refer to this as a _false positive_ change).

These false positives can easily end up being committed to the project, especially when the user has edited many files, for he/she might just add all changed files to the stage and commit without noticing it. These changed settings would then be picked up by other users' IDE after pulling the repository updates, if they happen to adopt the same IDE "__Save Setting to__" configuration. Because the IDE hides these end-of-file settings from the user, it's easy to forget about them or fail to notice their presence (unless one looks closely at the commit diff).

In this case, the solution is rather straightforward, for we only need to warn users (in `Contributing.md`) about the need to choose another option for "__Save Setting to__" — ideally, "__A common file project.cfg for every directory__" would be the best choice in collaborative projects, but "__The file &lt;filename&gt;.pb.cfg__" is also fine (and, of course, there's always "__Don't save anything__", but this might be too drastic and unnecessary).

## Git-ignoring .cfg Files

In any case, opting for the other two alternatives would still generate inside the repository tree one or more ever-changing files that Git would detect as being modified. For this reason, we ought to ad to `.gitignore` exclusion rules for these files:

```gitignore
project.cfg
*.pb.cfg
*.pbi.cfg
```

Currently, there are a various `project.cfg` in the repository:

    project.cfg
    DialogManager/project.cfg
    DialogManager/project.cfg
    PureBasicIDE/project.cfg
    PureBasicIDE/data/project.cfg
    PureBasicIDE/FormDesigner/project.cfg
    PureBasicIDE/sdk/project.cfg
    PureBasicIDE/tools/project.cfg

> __NOTE__ — the proposed exclusion rules are already in place in the[ `.gitignore` file](https://github.com/fantaisie-software/purebasic/blob/master/.gitignore#L6):
>
> ```gitignore
> ## ===============
> ## PureBasic files
[...]

PR #42, opened 2020-01-06
Check for IDE Settings in Travis Build

tajmone:
The firs of these two commits solves the problem of IDE Settings creeping into the repository files accidentally (as discussed in #15). In the end, I opted to use GREP to check for the presence of IDE setting in any PB sources (`*.pb`, `*.pbi`, `*.pbf`).

The second commit removes IDE settings from a number of sources that still had them (not sure if they were there before or if they creped back with some recent PR).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,11 @@ install:
   - npm install -g eclint
 
 script:
-  # ==============================================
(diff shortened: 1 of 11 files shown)


fantaisie-software:
Perfect

Issue #13, opened 2019-12-09
Add Pre-Compiled C Components


tajmone:
From the various comments on the [open sourcing announcement on the Forum], it seems clear that for many PB users the C compilation side of this project is either a deterrent or an obstacle to participation in the project.

I was thinking that it might be a good idea to add to the repository the pre-compiled binaries of the C components required by the PB IDE, so that end users don't have to deal with all the trouble of setting up MSVC, etc. Providing binaries for all the supported OSs and architectures would be ideal.

Although tracking binary files with Git usually isn't a good idea, this context might be an exception for various reasons.

1. These binaries are not going to be huge in size, and unless they change very often they won't bloat up the index.
2. There are already some exe files tracked by the repo (`DialogManager/PureBasic.exe`, `PureBasicIDE/tools/png2ico.exe`).
3. Having access to correctly pre-compiled binaries makes testing more reliable because everyone uses the same exact components. If these binaries come from the maintainers everyone feels more confident about the general setup.
4. Contributors could focus just on PureBasic code in this project, without having to worry about C compilation in order to build the IDE.
5. It would definitely broaden the audience of PB users getting involved in building the IDE from sources (even the curious one might give it at try).

That is, unless the C compilation process is interdependent with the PB sources, requiring recompilation each time (haven't had a chance to look into it close enough).

Alternatively, these binaries could be hosted somewhere on the WWW (including the project's Wiki) and download links provided.

On Windows it might be quite troublesome to setup all the required MSVC tools, especially for those who have already installed a more recent Win SDK and VS product (if I remember correctly, you can't install multiple SKDs on the same machine).

For example, I recently had to install the MS Visual C++ 2019 Build Tools and the Win 10 SDK in order to work with [Rust] lang. I think that this precludes installing the Win 7 SDK on this machine, so I wouldn't be able to replicate the exact same conditions in which the IDE is being developed so far. I was hoping to be able to avoid having to tackle with (yet another) virtual machine and installing even more tools, if possible, for my machine is already bloated by all the dependencies of various languages and tools (Ruby for Windows now requires installing [MSYS2] too). So, unless it's _really_ necessary to do so, I'd like to skip the C part (and I'm not a fan of C either, if I have to be honest).

[open sourcing announcement on the Forum]: https://www.purebasic.fr/english/viewtopic.php?p=545582
[MSYS2]: https://www.msys2.org/
[Rust]: https://www.rust-lang.org/

fantaisie-software:
[...]

PR #43, opened 2020-01-06
Fix misspellings

SicroAtGit:
I have executed the tool [`codespell`](https://github.com/codespell-project/codespell) on all files in the repository and carefully checked the fixes. But it would be good, if someone who has better knowledge of the English language would check the corrections again.


tajmone:
> But it would be good, if someone who has better knowledge of the English language would check the corrections again.

I would have loved to fix the misspellings and help with English in the comments, etc. The only thing which is holding me back is that **I can't build the IDE on my machine** (can't install the required version of MSVC, due to conflicts with my Rust setup) so I'd have to work in the dark, and I don't like it because I might break the code in the repo — it has already happened, with the `http://` to `https://` fix, which I didn't realize it broke some functionality in the IDE.

If someone could at least send me the pre-compiled binaries of the C modules of the project, for Windows, then I'd be able to test any changes to the code.

In a private email exchange with a Linux PB user I've learned that some people are not able to build the project under Linux either, and have the same problems regarding testing. 

We really need to solve this precompiled binaries issue (see #13) because is holding back some users from contributing (not only on Windows but also on Linux apparently). We can't reasonably ask for someone who'd like to help just us with fixing spelling issues to install the whole SDK and MSVS IDE. Besides, we need _official binaries_ to test changes to PB code against the same exact binaries that will be used in the official package.

tajmone:
## UK or US English?

@SicroAtGit, when fixing misspelling do you stick to UK English or USA English?

It might be a good idea if we all agreed on the same English standard, to prevent discrepancies — @fantaisie-software, any preferences here?

Most modern IDEs provide both dictionaries, so it should be easy for end users to just select the required English dictionary and use that as a reference.

kenmo-pb:
I read through the changes. Most fixes are in *comments* (harmless) but some are in *strings* (must be careful!) I see one problem:

- Please don't change `"Intervall"` to `"Interval"` in `Language.pb` and `Preferences.pb` because it will break compatibility with existing prefs files and translation files. (The misspelled string is used as a key, it's not just a harmless string displayed to the user.)

I think the PB team officially uses **US English**, the obvious example is that "Color" not "Colour" is used throughout the language.

tajmone:
> I read through the changes. Most fixes are in comments (harmless) but some are in strings (must be careful!) ...

We really should annotate those critical strings because it's so easy to overlook them — I also broke things with my PR #10 when I changed `http://` to `https://` in some places. A simple comment with a `** DON'T RENAME!! **` warning could have prevented this from happening.

> - Please don't change `"Intervall"` to `"Interval"` in `Language.pb` and `Preferences.pb` because it will break compatibility with existing prefs files and translation files. (The misspelled string is used as a key, it's not just a harmless string displayed to the user.)

Is that a critical settings? If it went lost due to renaming, what would the default value do? A possible workaround could be to map `"Intervall"` to `"Interval"`, so previous settings are not lost. The idea of having to be stuck forever with any critical keys that are misspelled is not cool.

Overall, I see getting rid of those misspellings as a significant enhancement to the overall quality of PB/SB. They are plenty and all over the place — and frankly, they look bad, and more so when they're found in settings or UI menus. Most editors today include automated spell checkers, and these misspellings light up the source files like a Christmas tree (as I'm writing this reply I can already see `"Intervall"` in red all over the page), so it's not like they're going by unnoticed even with non native speakers.

> I think the PB team officially uses US English, the obvious example is that "Color" not "Colour" is used throughout the language.

That's a good indicator. Consistency in terms of UK vs US English is not always easy nor possible to achieve in this field, due to some computer jargon being already consolidated in one variant or the other, e.g. API functions being immutably bound to one of the two, etc. Which is one more reason why we need an official statement as to which base dictionary to adopt.

kenmo-pb:
IMHO I would not reset a user's preference to fix a typo buried deep in an INI file. The other 99% of typos, that are in the UI and more visible, yes definitely.

> The idea of having to be stuck forever with any critical keys that are misspelled is not cool.

Agreed, it's why HTTP servers still misspell `Referrer` decades later :)  It's extra important not to have typos in strings exchanged between files or programs.

**Good news**: I noticed tonight, that `"Intervall"` is actually used in sections of code which are currently commented out. So it doesn't really matter. GitHub doesn't syntax highlight PureBasic so I did not notice.

**Bad news**: "Intervall" still exists as part of longer keys, such as `PurifierIntervall` and `GlobalIntervall` in `Languages.pb`. These were not fixed by the `codespell` tool. Also every IDE translation out there in the world contains these misspelled keys. So, again, I would recommend *not* changing these INI keys — and if we do, then map the old spelling to the new spelling, for compatibility, as you mentioned.

tajmone:
> GitHub doesn't syntax highlight PureBasic so I did not notice.

Yes, it's hard to catch these when there are many changed files involved. I use [BeyondCompare] for diffing, which allows to exclude comments from the diffs, a useful feature in these cases for it allows to view only code-significant changes in one go, working on a whole repository. I had created a PureBasic syntax for the tool, so it also provides some basic highlighting (the highlighting engine is not particularly smart nor friendly to use, but it's better than nothing).

I wanted to propose a PureBasic syntax for GitHub highlighter, but it won't be accepted unless the language scores a certain popularity rate.

> if we do, then map the old spelling to the new spelling, for compatibility, as you mentioned.

Is that feasible to accomplish or does it require lot of work?

[BeyondCompare]: https://scootersoftware.com/

kenmo-pb:
Using "Intervall" as an example... If only a few settings are renamed, I think we could handle them in-place:

    If ReadPreferenceString("Intervall", "") <> "" ; check if old spelling is present
      CPUUpdateInterval = ReadPreferenceLong("Intervall", 1000) ; old spelling
    Else
      CPUUpdateInterval = ReadPreferenceLong("Interval", 1000) ; new spelling
    EndIf

If we ever renamed many settings, we should probably handle them in bulk, like:

    ; Run this before reading all the preferences
    ForEach RenamedSetting()
      If ReadPreferenceString(RenamedSetting()\NewKey, "") = ""
        OldValue$ = ReadPreferenceString(RenamedSetting()\OldKey, "")
        If OldValue$ <> ""
          WritePreferenceString(RenamedSetting()\NewKey, OldValue$)
        EndIf
      EndIf
    Next

Another method is to check the saved `PrefsVersion` and read the file slightly differently depending on version.

Of course, this all applies to `Preferences.pb`, you'd have to do the same thing in `Language.pb` (so that renaming a language key doesn't cause a translated IDE string to be missed).

I think fixing the typos in code comments and UI strings can all be merged, but changing any INI keys should be a separate Pull Request because of this other work involved.

tajmone:
@kemno-pb, I totally agree with you: this PR should roll-back changes to misspelled setting keys, and just keep the fixes to the comments, so it can be safely merged without delays.

@SicroAtGit, are you OK with this?

I've opened Issue #44 to discuss whether it's worth or not to tackle with this (Fred's opinion, based on his experience on maintenance of similar issue) is vital to decide the faith of this. That would be the proper place to enter into details regarding how to work around eventual fixes, and the possible implications. I've also proposed adding some comments to protect special keys from accidental changes in the future, as well as using `[sic]` in comments that reference misspelled keys.

SicroAtGit:
@tajmone

Under my Manjaro Linux I can compile the IDE with [this shell script](https://github.com/tajmone/purebasic/blob/fix-ide-PB-contributing-PR/Linux-x64.sh). I probably can't help with the problems under Windows, because I have no experience with the C programming language and its development environments.

> when fixing misspelling do you stick to UK English or USA English?

I think it's US English.

---

@kenmo-pb @tajmone

The CPU monitor settings are a remnant of the earlier PB versions:

    16th December 2008 : Version 4.30
    ...
    Removed CPU monitor from the IDE, as all OS provide the same functionality

As you have already noticed, these settings are also commented out.

Earlier PB versions may not be so easy to restore, because probably a lot has changed, which prevents backwards compatibility. The only purpose of keeping these misspellings is to keep the documentation of obsolete features correct. Additionally, I would like to mention that keeping old code as comment block is bad programming style. Especially because we now use a version control system.

> A possible workaround could be to map "Intervall" to "Interval", so previous settings are not lost. The idea of having to be stuck forever with any critical keys that are misspelled is not cool.

These workarounds should still exist for all future PB versions, which is also not cool. Best we leave the misspellings that are in keys of settings, as @kenmo-pb prefers it, too. That way we save ourselves the unnecessary extra work.

---

@tajmone

> @SicroAtGit, are you OK with this?

I'm going to amend my commit so that it doesn't correct the misspellings in the keys of the settings. The other spelling corrections concern debug/log output, which probably isn't problematic, but I will look at the changes again tomorrow.

>  I've also proposed adding some comments to protect special keys from accidental changes in the future, as well as using [sic] in comments that reference misspelled keys.

That's a good idea.

SicroAtGit:
Okay, I amended the commit.

It now only corrects misspellings in comments, log/debug output and GUI texts.

With these changes I can compile and run the IDE without errors.

kenmo-pb:
@SicroAtGit Thanks for digging up that CPU change history. It was removed ~12 years ago! Yes it's probably time to delete that commented section.

Thanks for amending. Another reason I think key typos should be handled separately (now Issue #44 ) is that `codespell` did not catch the typos in multi-word keys like `NoIndependandColors`

PureBasicIDE/CompilerInterface.pb
```diff
@@ -219,7 +219,7 @@ CompilerIf #CompileWindows
           Language$ = Left(*Target\VersionField$[17], 4)
         EndIf
         
-        ; correct the version fields 1 and 2 if needed ( if wrongly formated they leed to a resource error)
+        ; correct the version fields 1 and 2 if needed ( if wrongly formatted they leed to a resource error)
```
kenmo-pb:
```suggestion
        ; correct the version fields 1 and 2 if needed ( if wrongly formatted they lead to a resource error)
```

PureBasicIDE/Explorer.pb
```diff
@@ -100,7 +100,7 @@ Procedure UpdateExplorerPatterns()
   ForEver
   
   
-  ; sanitise the selected patterns to avoid a crash when it is too hight (may heppen if you remove a tool!)
+  ; sanitise the selected patterns to avoid a crash when it is too high (may heppen if you remove a tool!)
```
kenmo-pb:
```suggestion
  ; sanitize the selected patterns to avoid a crash when it is too high (may happen if you remove a tool!)
```

PureBasicIDE/PureBasic.pb
```diff
@@ -5,8 +5,8 @@
 ;--------------------------------------------------------------------------------------------
 
 ; Change: Now *every* file in the source tree is included from this main file
-; This will reduce the number of XIncludeFile's in each file, and there is no more need to check dependancies.
-; Even platform specific files are always includeded. wether their code is compiled
+; This will reduce the number of XIncludeFile's in each file, and there is no more need to check dependencies.
+; Even platform specific files are always includeded. whether their code is compiled
```
kenmo-pb:
```suggestion
; Even platform specific files are always included. Whether their code is compiled
```

Window-x64.cmd
```diff
@@ -1,7 +1,7 @@
 @echo off
 
 :: Set the PureBasic home directory to a real x64 PureBasic installation (warning:
-:: when compilig, the IDE will be overwriten with the new one.
+:: when compilig, the IDE will be overwritten with the new one.
```
kenmo-pb:
```suggestion
:: when compiling, the IDE will be overwritten with the new one.
```

Window-x86.cmd
```diff
@@ -1,7 +1,7 @@
 @echo off
 
 :: Set the PureBasic home directory to a real x64 PureBasic installation (warning:
-:: when compilig, the IDE will be overwriten with the new one.
+:: when compilig, the IDE will be overwritten with the new one.
```
kenmo-pb:
```suggestion
:: when compiling, the IDE will be overwritten with the new one.
```

SicroAtGit:
@kenmo-pb 

Thanks for the additional fixes.

> `codespell` did not catch the typos in multi-word keys like `NoIndependandColors`

You can teach `codespell` this support via RegEx:
`$ codespell --regex "[A-Z]+[a-z'-]+|[a-z'-]+" .`

But then the automatic fix won't work anymore: `--write-changes` (write changes in place if possible)

> --regex REGEX
                        Regular expression which is used to find words. By default any alphanumeric character, the underscore, the hyphen, and the apostrophe is used to build words. **This
                        option cannot be specified together with --write-changes.**

(Extract from `codespell --help`; relevant text in bold)

---

@kenmo-pb @tajmone

I'll have a look at the other misspellings that now appear due to the RegEx variant in a few hours. I'll give feedback when I'm done. That way we avoid working on the same fixes.

SicroAtGit:
It took a little longer, but I'm done now!
Only additional misspellings in the comments were corrected.

I also corrected misspellings in variable names, function names and file names, but these corrections require intensive testing, so they are not included in this commit. For these fixes, I will create a new branch based on the master branch and create a new pull request.

SicroAtGit:
I better wait with the other pull-request (corrected misspellings in variable names, function names and file names), because in some places the same lines are changed. So there would be conflicts during the merge and if we are not careful when solving the conflicts, already existing fixes could be removed accidentally.

fantaisie-software:
Nice !

PR #45, opened 2020-01-15
Git compliance improvements/cleanup

kenmo-pb:
Here are some small changes which I think improve the repo:

- Set Git to ignore the generated `Build` subfolder
- Remove the hard-coded constant `#PureBasicPath` which doesn't seem to be used anymore
- Change `#BUILD_DIRECTORY` from a hard-coded absolute path to a relative path
- Set `PureBasicPath$` to `#PB_Compiler_Home` instead of a hard-coded path
- Prevent 2 debugger errors whenever `Scintilla.dll` is not found (can happen a lot while getting the repo up-and-running)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ ___*.*
 *.lnk
 *.url
 
+# Generated output files
+Build/
+
 ## ==================
 ## Binary Executables
 ## ==================
--- a/BUILD.md
+++ b/BUILD.md
@@ -50,7 +50,6 @@ PureBasic to open the "`PureBasic IDE.pbp`" project file and
 run it from PureBasic itself (be sure to adjust the constants in 'Compilers Options.../Contants')
 
 - The `#BUILD_DIRECTORY` constant must point to the Build/x64/ide/ folder created before by the 'make'
-- The `#PureBasicPath` constant must point to a full PureBasic installation.
 
 Don't hesitate to [drop a word] to improve this build guide, as right now it's very slim!
 
--- a/PureBasicIDE/PureBasic IDE.pbp
+++ b/PureBasicIDE/PureBasic IDE.pbp
@@ -358,11 +358,10 @@
       <format exe="console" cpu="0"/>
       <debugger custom="1" type="ide"/>
       <constants>
-        <constant value="#BUILD_DIRECTORY=C:\PureBasic\GitHub\purebasic\Build\x64\ide\" enable="1"/>
+        <constant value="#BUILD_DIRECTORY=..\Build\x64\ide\" enable="1"/>
         <constant value="#FredLocalCompile=1" enable="0"/>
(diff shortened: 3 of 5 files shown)


tajmone:
Thanks, this looks great!

I'm only perplex about:

> Set `PureBasicPath$` to `#PB_Compiler_Home` instead of a hard-coded path

Doesn't this affect how the build chain operates? i.e. trying to build the IDE directly in the PureBasic installation folder.

The default installation folder under Windows is a protected folder that requires admin privileges for write permission. Another problem is that under Windows that folder contains spaces in its path, which is incompatible with the Make part of the build chain.

I've already mentioned a few times that the current build strategy seems needlessly complex to handle, and that it would be much easier if the build chain would just take the default PureBasic installation folder an create a full copy inside the `Build/` dir (e.g. `PB570_x32_Win`, `PB570_x64_Win`, `PB570_x64_Mac`, etc.), and then work on the copy — instead of requiring custom defined paths or using the default install. This would also allow the same repository folder to be used for building under different OSs (e.g. a shared storage unit across different OSs, either on different machines or by using VMs on the same computer).

So I'm not sure how the above change affects the current build chain.

kenmo-pb:
@tajmone It should not affect the build chain. That `PureBasicPath$` I edited is not in a batch script or makefile, it's actually a runtime variable in the IDE code - it tells the IDE where to look for `Scintilla.dll` and some other resources - plus the section I changed is only compiled with the `#PB_Compiler_Debugger` ON. (But I would like Fred or somebody to double-check it.)

I 100% agree with your comments on the build paths! I find it strange that it overwrites the IDE in your PB installation folder. I would prefer my stable PB installation not be modified. And I have had Windows permission errors which prevent `make` from writing to `Program Files`, so I have edited my makefile to write to `Build/` instead.

Similarly, the `.pbp` project has 1 target with no Output Executable / icon / etc. defined. I think it makes sense to actually set up 6 targets, pre-defined as much as possible! Then you can selectively debug, Build All, etc.

- Windows x86
- Windows x64
- Mac x86 (still needed? isn't Apple killing 32-bit?)
- Mac x64
- Linux x86
- Linux x64

Of course this should all be tackled in another larger PR.

tajmone:
> @tajmone It should not affect the build chain.

Good to know. Let's hope it doesn't affect Fred's build environment either (there seems to be a delicate balance at play here, with the new needs of public users on the one side, and the original settings for the private project on the other).

> Of course this should all be tackled in another larger PR.

I agree. Also, this is something that should be approved by the project owner, since the main purpose of this repo is the production of the official PB packages. But, yes, definitely there are many problems on the end users' side right now, and the various proposals tend to cumulate too many new build scripts and script variations, only adding to the confusion.

fantaisie-software:
Very good !

PR #48, opened 2020-01-18
Remove user .cfg files

kenmo-pb:
Remove user-specific `.cfg` files to clean up the repo.

This is harmless because they only contain view settings (cursor position, folding status, etc.) and unneeded compiler settings (the `.pbp` file and `Makefile` contain everything necessary).


fantaisie-software:
Thanks, perfect !

PR #50, opened 2020-02-02
Fix display errors in window '#WINDOW_Form_Parent' of the FormDesigner

SicroAtGit:
PB 572 b1: IDE Formdesigner display problem setting parent
https://www.purebasic.fr/english/viewtopic.php?f=4&t=74526

- Height of the ComboBoxGadgets was too less
- Second TextGadget was not wide enough for the German text

Sorry, but only tested under Linux.


fantaisie-software:
Looks good. That said, we should convert this window into an XML dialog to avoid these kind of issues

SicroAtGit:
The changes in this pull request unfortunately do not lead to a satisfactory result under all operating systems. See the corresponding forum thread for more information.

I will create a new pull request tomorrow, in which the window is created using the dialog lib.

PR #47, opened 2020-01-18
Balance keywords to fix incorrect indentation

kenmo-pb:
There are some places in the code with "unbalanced" keywords (multiple opening keywords in `CompilerIf`/`CompilerElse` blocks, with only one closing keyword). This causes the keyword highlighter, "Go to matching keyword", and the auto-indenter to not work correctly.

This PR makes tiny code changes to "balance" all keywords. *Unfortunately this creates very large diffs because the indentation for the rest of the file is changed (fixed!)*

To better review the diffs, I recommend a viewer with an "Ignore Whitespace Changes" feature! (For example, WinMerge)

- I did not fix `EditHistory.pb` yet because I don't want to conflict with PR #34

PR #52, opened 2020-02-03
Balance keywords to fix incorrect indentation

kenmo-pb:
This is a new version of PR #47 , based off the latest `master` to solve conflicts, so I will repeat the description here:

> There are some places in the code with "unbalanced" keywords (multiple opening keywords in CompilerIf/CompilerElse blocks, with only one closing keyword). This causes the keyword highlighter, "Go to matching keyword", and the auto-indenter to not work correctly.
> 
> This PR makes tiny code changes to "balance" all keywords. Unfortunately this creates very large diffs because the indentation for the rest of the file is changed (fixed!)

Notes:

- Warning, this looks like a huge `diff` because of all the indentation changes. I recommend you "hide whitespace changes" in `git-gui` or use `?w=1` on GitHub, to review the changes. The "Split" view on GitHub also helps.
- This PR includes `EditHistory.pb` which #47  left out to avoid conflicts (it was already being modified by other Pull Requests)
- If we are serious about standardizing code style, and using the IDE Auto-formatter, then we don't want unbalanced keywords because they leave unfinished indentation toward the end of each file. (`EditHistory.pb` had a "leftover" indentation of 6 spaces)
- This is *not a functional IDE change*, it only cleans up the repo code! It only affects anyone reading or working on the IDE source.

### DIFF
--- a/PureBasicDebugger/Communication_Network.pb
+++ b/PureBasicDebugger/Communication_Network.pb
@@ -17,1409 +17,1411 @@ CompilerIf #CompileWindows
   Import "wsock32.lib"
   EndImport
   
-  ImportC #BUILD_DIRECTORY + "NetworkSupport.obj"
-  CompilerElse
+  #NetworkSupport_Library = "NetworkSupport.obj"
+CompilerElse
+  
+  #SOCKET_ERROR = -1
+  #INVALID_SOCKET = -1
+  
+  #NetworkSupport_Library = "NetworkSupport.o"
+CompilerEndIf
+
+ImportC #BUILD_DIRECTORY + #NetworkSupport_Library
+  
+  ; // Wrappers For the simple stuff For consistency
+  ; //
+  Network_Initialize()
+  Network_CreateSocket()
+  Network_CloseSocket(Socket)
+  
+  ; // Only initiates the connect.
+  ; // Use Network_ConnectSocketCheck() To wait For a connection/failure.
+  ; // Returns 0 on failure
+  ; //
+  Network_ConnectSocketStart(Socket, Hostname.p-ascii, Port)
+  
+  ; // Check the status of a connect()
+  ; //
+  ; // Returns:
+  ; //   0 = still pending
+  ; //   1 = connected
+  ; //   2 = connection failed
+  ; //
+  Network_ConnectSocketCheck(Socket)
+  
+  ; // Put the socket in listen mode (1 connection max)
+  ; // If Interface is empty, bind to all interfaces
+  ; //
+  Network_Listen(Socket, InterfaceName.p-ascii, Port)
+  
+  ; // Check If an incoming connection is made on a listening socket, does Not block
+  ; // returns new Socket Or SOCKET_ERROR (= no incoming connection)
+  ; //
+  Network_CheckAccept(Socket)
+  
+  ; // Check if data is ready For reading
+  ; //
+  ; // Returns:
+  ; //   0 = no Data
+  ; //   1 = Data available
+  ; //   2 = disconnect
+  Network_CheckData(Socket)
+  
+  ; // Ensure the Data is Read fully
+  ; // Returns true If ok, false If network disconnect
+  ; //
+  Network_ReceiveData(Socket, *Buffer, Size)
+  
+  ; // Blocking send of Data
+  ; // Returns true If ok, false If network disconnect
+  ; //
+  Network_SendData(Socket, *Buffer, Size)
+  
+  ; // Blocking send of string Data (ascii)
+  ; // Returns true If ok, false If network disconnect
+  ; //
+  Network_SendString(Socket, String.p-ascii)
+  
+EndImport
+
+ImportC ""
+  
+  ; Import of the AES functions (included in Cypher lib, linked because we use MD5 too)
+  ; (so we do not have to use the PB object system for thread safety)
+  ;
+  rijndael_set_key.l(*context, *key, bits)
+  
+  rijndael_ecb_decrypt.l(*context, *input, *output, length)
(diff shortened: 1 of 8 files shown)


tajmone:
I'm not sure of the benefits of manually formatting code over using the default IDE formatter. The whole idea of providing a standard way of linting is to avoid meaningless whitespace diffs, which this commit seems to introduce.

Although the IDE linter might not be perfect (and issues with stray indentation due to compiler directives falling across the boundaries of constructs are known) it still grants end users a standard way of reformatting the code, thus reducing meaningless changes in commits diffs.

Unfortunately, we can't validate whitespace in PB sources via EditorConfig, due to indented empty lines being produced by PB IDE.

So, lacking better solutions, asking all users to reformat the code with the PB IDE is currently the only way to grant formatting consistency. Of course, there are a few edge cases where linting goes wrong (compiler block directives cross the boundaries of constructs), but these are not the norm.

The only real solution to the problem would be to amend the IDE sources that deal with reformatting, to ensure that compiler directives don't lead astray indentation. Another alternative would be to create a independent linting script, and then find a way to impose it on the repository (the question being, in which language? ideally, end users shouldn't be asked to install an extra language just to submit contributions).

> Warning, this looks like a huge diff because of all the indentation changes. I recommend you "hide whitespace changes" in git-gui or use `?w=1` on GitHub, to review the changes. The "Split" view on GitHub also helps.

The proper way to handle this would be via `.gitattributes`, which also allow to control how diffs are handled, and these would be honoured by all Git related tools, including GitHub. I haven't used those settings before, I'm just aware of them, so I might look into them when I get some time (right now, I'm in and out of bed with fever).

Still, such settings only affect the visualization of diffs, but don't alter the fact that contributors will have to deal with all those whitespace changes — e.g. when this commits gets merged into `master`, anyone working on a branch will have to solve lots of conflicts to be able to rebase his/her changes.

Keep in mind that when the decision was taken to reformat all the sources using the IDE reformatter, we were already aware that some files would suffer from stray indentation (including some auto-generated files), but after discussing the issue the decision was that the benefits of enforceable standard code styles were greater than the readability price that would result in some places of the sources.

tajmone:
After looking at a proper diff of the changes (without the out-of-synch changes, as they appear in the default diff viewer), I seem to understand that you've solved the indentation unbalances by adding extra keywords to prevent crossing construct boundaries.

This wasn't clear to me in the first, from your statement:

> This is not a functional IDE change, it only cleans up the repo code! It only affects anyone reading or working on the IDE source.

(which seems to indicate manually fixing the indentation).

So, now the sources can be reformatted via the IDE and still obtain balanced indentations?

**PS:** I still think that the above proposal or revisiting the IDE formatter sources in order to handle correctly those edge cases makes sense.

kenmo-pb:
Sorry if the intention was not clear! More info:

I did not manually format anything — these 8 files are run through the IDE auto-formatter, but with minor code changes to "balance" opening and closing keywords. This keeps the indentation correct for the rest of the file. (Also it makes "Go to matching keyword" locate properly!)

`EditHistory.pb` was the worst case: 3 unbalanced sections caused the rest of the file to be indented by 6 spaces. That was the file I was working on for a PB diff implementation, and the weird indentation was annoying me.

This looks like a big diff, but really it is just correcting a subset of the big "Normalize PB sources" commit back in December.

I think we are early enough in the life of this GitHub project that it is worth fixing now. If not, then that's OK, the code works as-is.

tajmone:
Then it's good, well done — sorry for not grasping it immediately, but between the out-of-synch diff in the WebUI, and high fever slowing me down, I didn't get it right.

Yes, I agree, it's a good solution to make the code more readable, even though it adds some code redundancy here and there.

I was wondering how hard it might be to actually change the IDE formatter code so that it can handle compiler directives as special cases by either:

1. not affecting the indentation of what lies inside them (i.e. keeps indenting them according to the main code context), or
2. become aware of when a compiler directive block falls across the boundaries of a construct in the main code, to prevent loosing track of correct indentation.

Probably it might require a mixture of both approaches, depending on the context. 

Surely, it would have to cover many edge cases and might require extensive testing; but in the long run it might be worth it.

Also, I think that compiler directives themselves should never be indented, not even inside blocks, because they are a separate category of tokens. Eg:

``` purebasic
Select X
  Case 1
CompilerIf #PB_Compiler_ExecutableFormat = #PB_Compiler_Console
      PrintN("Hello world")
CompilerElse
      WriteStringN("Hello world")
CompilerEndIf
EndSelect
```


Regarding the diffs, I did look into Git's `.gitattributes` documentation trying to see how to change the rules for PB files diffs visualization, but I'm not sure if it can be done like I expected. It's all documented in a packed and unclear manner, so I might need to find some tutorials on the topic.

tajmone:
> We don't set the type explicitely in the IDE when it's the default type (`.i`).

Out of curiosity, @fantaisie-software why is this so? Is it a general best-practice rule (idiomatic PureBasic), or is it simply a convention of this specific project?

fantaisie-software:
We had history where we postfixed with '.l' everything when it was the default type. Then PB changed the default type to 'i' and the migration to 64-bit was a nightmare due to this. So the rule we like is to postfix only when you use a specific type (ie: not '.i'). As personal taste, I find it easier to read as well as it doesn't bring any more info. Alos the default type can't be changed anymore since a few PB version  with DefType so it's no more an issue

kenmo-pb:
@fantaisie-software Sure, I will push a small `.i` fix today.

kenmo-pb:
Removed `.i` from declarations.

PS. I originally miscounted, `EditHistory.pb` had *four* unbalanced keywords, so it was indented by *eight* extra spaces near the end of the file!

fantaisie-software:
Merged, thanks !

PR #54, opened 2020-02-06
New editor shortcut functions

kenmo-pb:
This PR adds 5 new user-configurable keyboard shortcuts, most are built-in Scintilla functions:

- Move selected lines up
- Move selected lines down
- Delete selected lines
- Duplicate selection
- Paste as comment

Notes / judgment calls:

- `Paste as comment` is specific to PB syntax, so it's a custom function. The other 4 are just calls to Scintilla functions. See comments in `UserInterface.pb`
- Based on Scintilla's defaults and other text editors, I gave 4 of the 5 default defined shortcuts, except `Delete selected lines`, which the user has to set.
- Another judgment call, I added `Paste as comment` to the menus under `Paste`. The other 4 are not in the menus, so they are grouped at the bottom of the preferences list (near `Shift comments to the right/left`).
- The new shortcut names are only in English! I did not translate any of them.

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -1046,6 +1046,9 @@ Enumeration 0
   #MENU_MoveLinesUp
   #MENU_MoveLinesDown
   
+  #MENU_DeleteLines
+  #MENU_DuplicateSelection
+  
   #MENU_AutoComplete
   #MENU_AutoComplete_OK ; can now have a custom shortcut too
   #MENU_AutoComplete_Abort
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -944,6 +944,7 @@ Enumeration 0
   #MENU_Cut
   #MENU_Copy
   #MENU_Paste
+  #MENU_PasteAsComment
   #MENU_CommentSelection
   #MENU_UnCommentSelection
   #MENU_AutoIndent
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -1043,6 +1043,9 @@ Enumeration 0
   #MENU_SelectBlock
   #MENU_DeselectBlock
   
+  #MENU_MoveLinesUp
+  #MENU_MoveLinesDown
+  
   #MENU_AutoComplete
   #MENU_AutoComplete_OK ; can now have a custom shortcut too
   #MENU_AutoComplete_Abort
--- a/PureBasicIDE/Declarations.pb
+++ b/PureBasicIDE/Declarations.pb
@@ -155,6 +155,7 @@ Declare Redo()
 Declare Cut()
 Declare Copy()
 Declare Paste()
+Declare PasteAsComment()
 Declare SelectAll()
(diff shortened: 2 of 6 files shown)


pb-null:
Here is the german translation for catalogs/deutsch/editor.catalog
(which isn't in the repo unfortunately? Maybe they are different per OS, I'm on Linux, where an & sign is used for menu shortcuts)
These will create duplicate menu shortcut key assignments. Maybe it's better to remove the &s in the new entries(?)
`
[after line 49, "Paste             = &Einfügen"]
PasteComment      = A&ls Kommentar einfügen

[after line 955 (now 956) "DeselectBlock       = Auf vorherige Code-Blockauswahl zurückkehren"]
MoveLinesUp         = Ausgewählte Zeilen nach &oben verschieben
MoveLinesDown       = Ausgewählte Zeilen nach &unten verschieben
DeleteLines         = Ausgewählte Zeilen &entfernen
DuplicateSelection  = Auswahl &duplizieren

`

kenmo-pb:
That was fast, thanks!

I think the translation files (.catalog) would be a very good thing to maintain here, and it would create the opportunity for new contributed languages!

`&` creates an underline/shortcut on Windows too, but not on Mac I believe.  
Notice I did not include `&` in my English names. That's because 4 of them I did not put in the menus (so they're only displayed in the Preferences window).  
`Paste as comment` I put in the menu, but I didn't give it a `&`. I can give it `&s`, because `s` seems to be the only available letter in the English `Edit` menu.

In fact, the `Edit` menu already has conflicting shortcuts: `I&nsert comments` and `Find &Next`, `&Paste` and `Find &Previous`. `Format indentation` has no `&`

fantaisie-software:
I agree for the catalogs, it could be put here soon. Good work on this !

kenmo-pb:
I noticed in `ScintillaHilightning.pb` that `Ctrl+D` is treated as a special case (a Scintilla default shortcut which is *not* removed – that's why it could be used despite not being an officially configurable PB IDE shortcut).

Should I remove that special treatment, now that `Duplicate Line` has been accepted as a new, configurable IDE shortcut?

    For c = 'A' To 'Z'
      If c <> 'D' ; do not remove CTRL+D
        SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_CTRL) << 16), 0)
      EndIf
      SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_ALT) << 16), 0)
      SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_SHIFT|#SCMOD_CTRL) << 16), 0)
      SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_SHIFT|#SCMOD_ALT) << 16), 0)
      SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_ALT|#SCMOD_CTRL) << 16), 0)
      SendEditorMessage(#SCI_CLEARCMDKEY, c | ((#SCMOD_SHIFT|#SCMOD_CTRL|#SCMOD_ALT) << 16), 0)
    Next c

Issue #55, opened 2020-02-07
More spelling issues


kpeters58:
Nice work so far!

```
PureBasicDebugger/DebuggerGUI.pb  
; checks if an event is coming from a debugger window and dispactes it
```
That should be 'dispatches' it. I believe I also saw instances of 'childs' instead of 'children'

SicroAtGit:
The pull request #56 fixes your mentioned misspellings :-)

SicroAtGit:
I think this issue can be closed.

Issue #44, opened 2020-01-09
Typos in Settings Keys


tajmone:
PR #43 (in an attempt to fix all misspellings) revealed that there are some misspelled strings paired to IDE Settings keys. Fixing their spelling would break backward compatibility with user settings.

We should:

- [ ] __Add a warning comment__ next to those strings to prevent accidental changes (in the meantime).
- [ ] In comments that mention misspelled strings/keys, add a "`[sic]`" next to the misspelled word to indicate awareness of its misspelling status and that it should not be amended.
- [ ] Decide whether it's worth the effort to fix the spelling of these settings key.

There are possible workarounds to this, like introducing a double-mapping standard to handle the old misspelled keys, but they require adding some _ad hoc_ code (see [@kenmo-pb proposal] at PR #43).

Whether it's decided in favour or against tackling with similar fixes, this discussion might lead to a general rule regarding this type of interventions, which would be worth adding to the Contributors Guidelines (to prevent the same proposal coming up over and over again). 

[@kenmo-pb proposal]: https://github.com/fantaisie-software/purebasic/pull/43#issuecomment-572576399

kenmo-pb:
@tajmone Good summary. Here are the misspelled keys I have found (but there could be more).

    NoIndependantColors
    NoIndependandColors
    AllreadyUsed
    PurifierIntervall
    GlobalIntervall
    LocalIntervall
    StringIntervall
    DynamicIntervall

Most of these are not in the *Preferences* but in the *Language* files, where it would be even *worse* to suddenly change keys.

If a preference key changes, OK, the default value will be used instead. The next save will fix the key.
*Actually - I realize right now - this might be a problem for users who switch back and forth between PB versions such as the LTS versions!*

If a language key changes, that's bad, because some strings won't be translated, and the IDE does not re-save and fix the keys like a `prefs` file!

Let's wait for Fred's opinion. I think these are very low priority. (Also we still need a development branch so we're not merging potential problems into `master`!)

tajmone:
> Actually - I realize right now - this might be a problem for users who switch back and forth between PB versions such as the LTS versions!

In theory, users should be using the latest IDE to handle all projects since the IDE allows you to choose which compiler version to use, just like you can use the x64 IDE to compile 32-bit binaries (which is usually what you do anyway in order to compile the same project for both architectures using a single IDE session). At least, that's what I usually do. It only requires adding all the available PB packages to the preferences file via the IDE.

[...]

PR #56, opened 2020-02-08
Fix misspellings (partly unsafe)

SicroAtGit:
As I mentioned in [my other pull request](https://github.com/fantaisie-software/purebasic/pull/43#issuecomment-573397573) with the same topic, I found some more misspellings.

---
ToDo:
- [x] Leave misspellings in the setting keys (#44)
- [x] Leave misspellings in the language keys
- [x] Add a note that misspellings should be kept in the two places above
- Fix more misspellings (only a few that were reported as issues)
  - [x] #55
- [x] Extensive testing to ensure that nothing is broken by the changes


kenmo-pb:
Hi @SicroAtGit , some thoughts:

1. Agreed, let's not change settings keys, as discussed with Fred
2. Similarly, let's not change language keys! For example `AllreadyUsed -> AlreadyUsed`. If we change that key, then the translation won't be found when looking in language files that already exist.
3. We must be very careful about renaming all these variables without having `EnableExplicit` on!  :warning: This contains *a lot* of renames that could break something...

SicroAtGit:
> 2. Similarly, let's not change language keys! For example `AllreadyUsed -> AlreadyUsed`. If we change that key, then the translation won't be found when looking in language files that already exist.

You're right, that should also remain untouched. I added it to the to-do list.

> 3. We must be very careful about renaming all these variables without having `EnableExplicit` on!  warning This contains _a lot_ of renames that could break something...

Yes, that's why I marked it as unsafe.
When the changes are complete and everyone agrees with the changes, I will do a very thorough review.

SicroAtGit:
Under Linux the compilation runs without errors.

When I change the language, I can't find any texts in the GUI that keep the default language (English).

Everything looks ok for me now.

PR #57, opened 2020-02-08
Check PB Sources for Trailing WS

tajmone:
Update `validate.sh` to check for trailing whitespace (tabs or spaces) in lines that contain at least one non-whitespace character (i.e. don't consider indentation-only lines as trailing whitespace, because the PB IDE formatter allows them).

Now all PB sources with trailing space will fail the Travis CI checks.

I've tested it live with the latest merge of PR #53, which contained some trailing spaces (as pointed out by @SicroAtGit), and it correctly failed the build listing all PB files that didn't meet the requirements:

https://travis-ci.com/tajmone/purebasic/builds/148020124

### DIFF
--- a/PureBasicDebugger/MemoryViewer.pb
+++ b/PureBasicDebugger/MemoryViewer.pb
@@ -5,76 +5,76 @@
 ;--------------------------------------------------------------------------------------------
 Global Dim MemoryViewer_Chars.s(31)
 
-#MEMORY_VIEW_TABLE_DATA_DEC = 0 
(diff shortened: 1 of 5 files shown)


tajmone:
I didn't squash the two commits into a single one just because I wanted to keep a commit that failed Travis CI to check that the log error was looking fine and the script also worked on Linux (tested on Bash for Windows).

 But, if you prefer I can squash them into a single commit (even though I see nothing wrong with keeping them split, since they deal with independent steps, albeit strictly related).

PR #64, opened 2020-02-29
Fix re-enable buttons bug in window 'Find in Files' (#WINDOW_Grep)

SicroAtGit:
If the output window (#WINDOW_GrepOutput) was closed during a running
search, it was still possible to abort the search with a click on the
Stop button, but the Start and Cancel button was then not re-enabled.

https://www.purebasic.fr/english/viewtopic.php?f=4&t=71539


fantaisie-software:
Thanks for the fix !

PR #66, opened 2020-03-03
Fix selection of binary numbers

SicroAtGit:
:warning: Resolve the merge conflict carefully to avoid reversing the changes of the other PR (#67).

---

When binary numbers were selected by double-clicking, the prefix (%) of the binary numbers was not also selected.

When a binary number including the prefix (%) was selected, further occurrences of this binary number were not highlighted in the code.

Side effect:
If the character '%' is used as an operator in the current case and the variables and numbers are directly adjacent to the operator (without spaces), the variables/numbers directly adjacent and the operator are selected.

Bug in IDE - double click to select problem
https://www.purebasic.fr/english/viewtopic.php?f=24&t=74723

double click select binary value including %
https://www.purebasic.fr/english/viewtopic.php?f=3&t=74741

### DIFF
--- a/PureBasicIDE/ScintillaHighlighting.pb
+++ b/PureBasicIDE/ScintillaHighlighting.pb
@@ -3349,7 +3349,7 @@ CompilerIf #CompileWindows | #CompileLinux | #CompileMac
     ; add *@$# to the word characters, so they get included in the selection
     ; when you double-click a word (to so select constants/variables easily)
     
-    WordChars$ = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$@#*"
+    WordChars$ = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$@#*%"
     For k = 192 To 255
       WordChars$+Chr(k) ; For ASCII mode, to have "é, à" etc. (https://www.purebasic.fr/english/viewtopic.php?f=4&t=57421)
     Next


t-harter:
Not sure if this is the best idea. Use of the binary number notation is not that frequent in my observation. Use of % as the modulo operator is much more prevalent. This change would create many false positives where the behavior actually gets worse.

tajmone:
> This change would create many false positives where the behavior actually gets worse.

I've experienced this with syntax highlighters for PB, if the modulo operator is not space-separated from the following number it gets captured as a binary number. Same problem with editors syntaxes.

There's no perfect way to get around this except via a full parser that's able to contextualize the tokens, and since adding a space after the modulo is not mandatory (indeed, it's common practice not to add spaces after operators) it's an unavoidable edge case.

But in some editor syntaxes I managed to work around this by ensuring that binary numbers only contain zeros and one (i.e. via RegEx), if not then consider the "%" as the modulo operator. Of course, edge cases like "% 10" would still slip through if the space is omitted.

Ideally, end users would add a space after the modulo, to make the code more readable and less ambiguous. Other than that, not much can be done since it's valid syntax. 

Having said that, I'm in favor for these fixes for they seem to make the editor more consistent. Hopefully, end users will add a space after the modulo, to avoid ambiguity and get the best out of the editor.

kenmo-pb:
I don't have a strong opinion either way. It's a low-impact feature (specifically double-click-selecting a binary literal).

But for consistency, I agree with the Pull Request. When people use the modulo operator, I think they *usually* include spaces, so it won't negatively affect them. But when you write binary literals, you *never* include spaces so it will *always* positively affect you (easier selection and occurrence highlighting).

Besides, the selecting-too-much argument is weakened because the `*` character is already included! So multiplications will be selected if you don't include spaces, and those are way more prevalent than modulo. (eg. `speed*time` is completely selected on double-clicking).

kenmo-pb:
You can only do so much with `SCI_SETWORDCHARS`. It's not "intelligent".  
Maybe the best solution is to catch double-clicks and perform some custom PB selection logic? But then you might interfere with Scintilla's double-click-drag and triple-click features...

*Slightly off topic*  
More important in my opinion is to *remove* the `@` character. When you double-click a variable name, I don't think you usually want to include the Address Of prefix. (After all, it's truly an operator, and other operators aren't included.) Plus when you double-click-select `@variable` it does not highlight other occurrences of `variable`.

fantaisie-software:
I'm a bit mixed about this PR, as we can't imply than every coders will put space around modulo, and as Timo said, use of literal binary is very very small compared to modulo.

t-harter:
> Having said that, I'm in favor for these fixes for they seem to make the editor more consistent. Hopefully, end users will add a space after the modulo, to avoid ambiguity and get the best out of the editor.

Making changes for "consistency" and forcing users to adjust the way they write their code to comply with the IDE behavior is not a good way to keep up user satisfaction. Especially when you change how things have worked for years already. The consistent thing to do here would be to keep the behavior as it is. 

> More important in my opinion is to remove the @ character. When you double-click a variable name, I don't think you usually want to include the Address Of prefix.

That is a matter of opinion also. You can't expect everybody to agree with that. For them, your "fix" will break the behavior they are already used to.

kenmo-pb:
> The consistent thing to do here would be to keep the behavior as it is.

Of course when Tajmone said "consistent" he meant keeping behaviors consistent with each other, not consistent with their past behavior.  
The `%` char is not included, but the `@` `*` chars are, even though `*` is susceptible to similar false positives (selecting multiplication statements).

So, maybe the best idea is what Sicro said: make it a setting, and default to the existing behavior. 👍  Some possibilities:

1. Add some boolean settings like `_ Include @ character when selecting words`
2. Make the word character string a user-editable setting in the IDE
3. Make the word character string a user-editable setting only in the INI file, for advanced users
4. ???

t-harter:
% vs * is not the same thing IMHO because pointers are a lot more in use compared to the binary number notation. The false positives about * as multiplication were well known (even some bugreports about it) but we simply had no better option.

I think a config option is a good compromise. In this case I would even suggest to go with a consistent default (i.e. add * and % but not @) which is better for new users. People who want the old behavior back can easily switch it back if they want. 

I suggest to not make the whole `WordChars$` string a config option because I don't expect anybody wants to seriously remove "A" from the selection so it only adds unnecessary stuff. I suggest to only add an option for the special characters "$@#*%" and hard-code the rest. The "Editor -> Editing" prefs page has lots of space left...

kenmo-pb:
> I suggest to not make the whole `WordChars$` string a config option because I don't expect anybody wants to seriously remove "A" from the selection so it only adds unnecessary stuff. I suggest to only add an option for the special characters "$@#*%" and hard-code the rest. The "Editor -> Editing" prefs page has lots of space left...

Yes, always include `A-Z` `a-z` `0-9`, and make `_$@#*` a user-customizable string... I like that!

SicroAtGit:
@t-harter 
> In this case I would even suggest to go with a consistent default (i.e. add * and % but not @) which is better for new users. People who want the old behavior back can easily switch it back if they want.

I'm glad you accept my PRs as the new default setting. Thanks.

fantaisie-software:
Should we close this PR, as a the new one from kenmo has been pushed ?

PR #67, opened 2020-03-04
Fix selection of words with the address-of-operator '@'

SicroAtGit:
:warning: Resolve the merge conflict carefully to avoid reversing the changes of the other PR (#66).

---

If a word was selected, the other occurrences of this word in the code were only highlighted if they were not directly adjacent to the address-of-operator (with spaces after the address-of-operator).

IDE - Highlighting selected Variables
https://www.purebasic.fr/english/viewtopic.php?f=4&t=69291

### DIFF
--- a/PureBasicIDE/ScintillaHighlighting.pb
+++ b/PureBasicIDE/ScintillaHighlighting.pb
@@ -3349,7 +3349,7 @@ CompilerIf #CompileWindows | #CompileLinux | #CompileMac
     ; add *@$# to the word characters, so they get included in the selection
     ; when you double-click a word (to so select constants/variables easily)
     
-    WordChars$ = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$@#*"
+    WordChars$ = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$#*"
     For k = 192 To 255
       WordChars$+Chr(k) ; For ASCII mode, to have "é, à" etc. (https://www.purebasic.fr/english/viewtopic.php?f=4&t=57421)
     Next


t-harter:
I think it is also very useful to be able to quickly select a variable including the preceding @ for copy & paste. This change breaks that.

Josh-Pb:
@t-harter 
I don't think so.  The current behavior breaks the rules. When double-clicking on a variable, the operator is not included. Why should this be different for the '@' address operator? Unlike '*', the '@' is not part of the variable name.

More important is that when highlighting a variable, it is marked, no matter if there is an '@' in front of it or not. The current behavior has often driven me crazy when I have exchanged a variable, but not the '@' variables, because they are not considered in highlighting.

kenmo-pb:
> More important is that when highlighting a variable, it is marked, no matter if there is an '@' in front of it or not. The current behavior has often driven me crazy when I have exchanged a variable, but not the '@' variables, because they are not considered in highlighting.

I agree with this, but as Timo said in the other PR, it's a matter of opinion. So maybe a user setting is best. The simplest to implement would be a 'hidden' setting in the prefs file (not in the GUI) but I don't think the IDE typically does it that way.

@SicroAtGit Maybe I'm wrong, but by submitting these as two separate PRs (add `%` and remove `@`) aren't you creating a conflict by changing the same line of code in two different ways? A simple conflict to resolve of course :)

A workaround for the current IDE is: open the IDE in a hex editor, search for the word chars string (UTF-16 encoded), replace `@` with `%` in-place. 👍 Silly but it works.

tajmone:
> I agree with this, but as Timo said in the other PR, it's a matter of opinion. 

True, I had to face this dilemma wile working on my PB syntax for Sublime Text, having to decide whether to include symbols like the `#` for constants as part of the parsed token or not. The recommended way for most editors is to include such symbols as part of the token, which restricts selections (and symbols look up) to the identifiers that carry the same symbol.

Personally, I'd favor IDE consistency in behavior, although I admit that in PB is useful to be able to select all occurrences of similar named variables when one wants to do dome global renaming (i.e. pointers, constants, etc., which share a same name).

Josh-Pb:
Sorry Tajmone,

you're mixing things up that have nothing to do with each other:

- The '*' for pointers and the '#' for constants **are part of the identifier** and must be included in highlighting. *abc, #abc and abc are three different identifiers that have absolutely nothing to do with each other.

- The @ is an address operator and **is not part of the identifier**.

Although I appreciate Timo's knowledge and dedication very much, it is incomprehensible for me why a single operator is picked out and this one is highlithed with the variable name. Besides the fact that this behaviour is extremely annoying, it is also completely inconsistent. The '?' as second address operator, or a unary minus are also not taken with the identifier.

The argument that this is convenient for copy and paste is also incomprehensible to me. When I set or change a name, it can occur with or without an @. So this behaviour is completely counterproductive for me in this case, too.

SicroAtGit:
> @SicroAtGit Maybe I'm wrong, but by submitting these as two separate PRs (add `%` and remove `@`) aren't you creating a conflict by changing the same line of code in two different ways? A simple conflict to resolve of course :)

Yes, that's right. But unlike the "Fix misspellings" PRs (#43, #56) the change in this PR and the other one (#66) is very small and the conflict is easy to resolve. Maybe one PR for it both would have been better here, I'm not sure. If I have the time, I can rebase the other PR's commit so that the merge can be done without conflict.

In the first "Fix misspellings" PR I also pointed out the conflict problem: https://github.com/fantaisie-software/purebasic/pull/43#issuecomment-573397573

> So maybe a user setting is best.

Yes, I think so, too.

tajmone:
@Josh-Pb:

> you're mixing things up that have nothing to do with each other:

No, I didn't. I was mentioning a _similar_ issue with other editors syntaxes; and anyhow it's related, because the general question is whether to allow selecting only the identifier-part of a variable, constant or pointer (or `@` prefixed identifier) in order to quickly rename all same-named identifiers (even if different tokens), or whether to always include the preceding symbol in the selection.

Some user will find useful the above feature, in case they change their mind regarding some identifiers convention (e.g. `RE_...`, `#RE_...` and `*RE_...`) and wish for a quick way to batch rename them. Others will prefer to exact matches only. Hard to say which is better, but it seems to me that most editor go for the exact match (i.e. including the symbol in the selection).

Either way, the IDE should be consistent in how it handles selecting identifiers preceded by a symbol (be it `#`, `@` or `*`).

Josh-Pb:
> Either way, the IDE should be consistent in how it handles selecting identifiers preceded by a symbol (be it `#`, `@` or `*`).

Sorry, you just didn't get the difference. You're mixing up completely different things that have absolutely nothing to do with each other.

Please read the two points in my last posting where I explained the difference and try to understand it.

tajmone:
> Sorry, but this sentence alone shows that you didn't get it.

No, it just shows that I have a different opinion on the matter. Let's leave at that.

fantaisie-software:
Should we close this PR, as a the new one from kenmo has been pushed ?

PR #68, opened 2020-03-05
Fix highlight of matching word at end of file

kenmo-pb:
This fixes a minor bug in the "Repeated Selections Background" feature.

If you select a word (such as `EndIf`) it highlights all other occurrences. But if it occurs at the *very end of the file* (no newline) that one is not highlighted.

This fixes it by changing the boundary of the `While` loop from a `<` to `<=` check.

For example, if the selection length is 5 (`EndIf`) and the source length is 100 bytes, the final check of the loop should be bytes `95–99` (0-based). Previously it was stopping the loop one byte before that.


fantaisie-software:
Thanks !

PR #71, opened 2020-03-06
Make Scintilla word chars a user preference

kenmo-pb:
This PR allows the user to customize what characters Scintilla considers "word characters", for example when double-clicking to select a variable name. It also affects Find/Replace > "Whole words only" behavior.

Notes:

- This is a compromise of PR #66 and and #67 , see those discussions
- I left the defaults as-is (`$#*@`). Should we add `%` or remove `@`?
- I added it to the Preferences under Editor > Editing per @t-harter 
- I labeled it "Word characters" but we could use a better description
- This requires you to run `make` again because the Preferences dialog XML changed
- I moved the `SCI_SETWORDCHARS` call into a procedure, so it can be called at Scintilla creation OR applying changed preferences
- Because it's user-customizable now, I added some basic checks (disallow whitespace characters, interpret an empty setting string as "use defaults")
- `A-Z` `a-z` `0-9` `_` are always automatically included
- The IDE also includes characters 192-255 (existing behavior), is this still correct?

### DIFF
--- a/PureBasicIDE/CodeViewer.pb
+++ b/PureBasicIDE/CodeViewer.pb
@@ -214,7 +214,7 @@ EndProcedure
 Procedure InitCodeViewer(Gadget, LineNumbers)
   ScintillaSendMessage(Gadget, #SCI_CLEARCMDKEY, #SCK_TAB) ; to enable the window shortcuts
   ScintillaSendMessage(Gadget, #SCI_CLEARCMDKEY, #SCK_RETURN)
-  ScintillaSendMessage(Gadget, #SCI_SETWORDCHARS, 0, ToAscii("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$@#.*")) ; for simpler selecting
+  ApplyWordChars(Gadget)
   ScintillaSendMessage(Gadget, #SCI_SETMARGINWIDTHN, 1, 0)
   
   If LineNumbers
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -347,6 +347,7 @@ Enumeration 1 ; 0 is reserved for uninitialized #PB_Any
   #GADGET_Preferences_EnableCaseCorrection
   #GADGET_Preferences_EnableLineNumbers
   ;  #GADGET_Preferences_EnableMarkers
+  #GADGET_Preferences_ExtraWordChars
   #GADGET_Preferences_ShowWhiteSpace
   #GADGET_Preferences_ShowIndentGuides
   #GADGET_Preferences_EnableBraceMatch
@@ -1939,6 +1940,8 @@ EndEnumeration
 
 #ZOOM_Default = 0
 
+#WORDCHARS_Default = "$#*%"
+
 
 
 ;
@@ -2444,6 +2447,7 @@ Global EnableBraceMatch, EnableKeywordMatch, ShowWhiteSpace, ShowIndentGuides, M
 Global FormVariable, FormVariableCaption, FormGrid, FormGridSize, FormEventProcedure, FormSkin, FormSkinVersion
 Global FilesPanelMultiline, FilesPanelCloseButtons, FilesPanelNewButton
 Global CurrentZoom, SynchronizingZoom
+Global ExtraWordChars$
 
 ; Dialog Window data
 ;
--- a/PureBasicIDE/Declarations.pb
+++ b/PureBasicIDE/Declarations.pb
@@ -131,6 +131,7 @@ Declare UpdateHighlighting()                                   ; highlight every
 Declare StreamTextIn(*Buffer, Length)                          ; put the given buffer into the current source
 Declare StreamTextOut(*Buffer, Length)                         ; get the contents of the current source into the buffer
 Declare GetSourceLength()                                      ; get the source length in bytes
+Declare ApplyWordChars(Gadget = #PB_All)                       ; update the word chars for a specific ScintillaGadget, or all sources if #PB_All
 Declare RefreshEditorGadget()                                  ; redraw the editorgadget
 Declare ChangeActiveLine(Line, TopOffset)                      ; change the currently active line
 Declare SetBackgroundColor(Gadget = -1)                        ; set the bacground color fur this source (also for linenumbers)
--- a/PureBasicIDE/DiffWindow.pb
+++ b/PureBasicIDE/DiffWindow.pb
@@ -682,7 +682,7 @@ Procedure OpenDiffWindow()
       For Gadget = #GADGET_Diff_File1 To #GADGET_Diff_File2
         ScintillaSendMessage(Gadget, #SCI_CLEARCMDKEY, #SCK_TAB) ; to enable the window shortcuts
         ScintillaSendMessage(Gadget, #SCI_CLEARCMDKEY, #SCK_RETURN)
-        ScintillaSendMessage(Gadget, #SCI_SETWORDCHARS, 0, ToAscii("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$@#.*")) ; for simpler selecting
+        ApplyWordChars(Gadget)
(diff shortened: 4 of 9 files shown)


kenmo-pb:
Here's a preview of how it looks:

image

tajmone:
> I labeled it "Word characters" but we could use a better description

"Matched identifiers prefixes"?

kenmo-pb:
> "Matched identifiers prefixes"?

But `$` is a suffix 😁

SicroAtGit:
> I left the defaults as-is (`$#*@`). Should we add `%` or remove `@`?

Please add `%` and remove `@` — https://github.com/fantaisie-software/purebasic/pull/66#issuecomment-595924512

>     * I labeled it "Word characters" but we could use a better description

Special word characters?
And add a tooltip to the TextGadget that lists the hardcoded word characters.

tajmone:
> But $ is a suffix 

True.

"Matched identifiers prefixes/suffixes"?

It has to be short and intuitive.

t-harter:
There are more uses of SCI_SETWORDCHARS which should be adjusted to behave the same way:
https://github.com/fantaisie-software/purebasic/blob/00584883b19726dbe4cdab6a607830b20554654a/PureBasicIDE/CodeViewer.pb#L217
https://github.com/fantaisie-software/purebasic/blob/00584883b19726dbe4cdab6a607830b20554654a/PureBasicIDE/DiffWindow.pb#L685

Interestingly, those contain the "." character too which does make some sense as well. Thoughts? :)

As for the prefs name. I think it should hint a bit better what its primary use is. How about "Extra word characters (selection)" ?

> The IDE also includes characters 192-255 (existing behavior), is this still correct?

Its for french accents and german umlauts which can occur inside words in comments/strings. This behavior should remain.

tajmone:
The technical term for both prefix and suffix is **affix** (but it's less intuitive than using **prefix/suffix**).

I agree with @t-harter:

> As for the prefs name. I think it should hint a bit better what its primary use is.

The challenge here is delivering a description that is both intuitive and concise in a small space.

- Positioning the input box below the description would allow a little more room to the description text (keep in mind that other locales translations might require a longer text that English). With more room for the text, we could make the option more clear.

- Adding a  tool-tip with  a longer description might help clarifying what the option does, without need to read the Help file. So we might be looking for two sentences here: 

1. the brief text for the option label
2. a longer text for the tool-tip pop-up text.

It's also worth noting that we should open a separate Issue for documenting this feature in the PB Help file. Since the documentation is not part of this repository, all we can do for now is to create a reminder for new (or changed) IDE features that need to be documented (and maybe paste a draft), so that the documentation maintainers can keep track of them — otherwise the documentation will end up being out-of-sync with the actual IDE features.

kenmo-pb:
> Please add `%` and remove `@` — [#66 (comment)](https://github.com/fantaisie-software/purebasic/pull/66#issuecomment-595924512)

@SicroAtGit OK

> There are more uses of SCI_SETWORDCHARS which should be adjusted to behave the same way

@t-harter Ahh, usually I am good about searching the project. This time I rushed it.

> Interestingly, those contain the "." character too which does make some sense as well.

That makes sense for selecting numbers, because `.` is not an operator. I can add it. Of course `-` and `+` can be part of numbers too, but I think we agree the `-` and `+` operators are more important, so let's *not* select those by default?

> Positioning the input box below the description would allow a little more room to the description text

@tajmone Good idea, I'll change it.

About the tooltip... do any other IDE preferences use tooltips? Would it be weird if just one setting has a tooltip... @t-harter ? (I don't think the tooltip should list all other word chars, because it will be over 100 characters long.)

kenmo-pb:
Actually, I'm having second thoughts about including `.` and `%` as defaults now.

My biggest concern about the word chars is that somebody pointed out it affects Find/Replace "Whole words only". This is dangerous especially *without EnableExplicit*.

Consider this example:

    this = 22
    that = 5
    
    result = this%that
    Debug result ; should be 2

If you rename the variable `this` by "Whole words only", then having `%` in the word chars will *not* rename `this` in the modulo operation. Causing `result` to be `0`. Of course EnableExplicit would help here.

And if we include `.` then selecting/replacing a variable called `MyInt` will not include anywhere that it's written `MyInt.i` including the definition.

tajmone:
@kenmo-pb:

> About the tooltip... do any other IDE preferences use tooltips? Would it be weird if just one setting has a tooltip... @t-harter ?

Personally, I think that if tooltips make an option easier to understand they should be employed.

But one think that we need to consider (and I didn't) is how this might affect translations of the IDE. I'm not sure if the translation files would cover tool-tips too.

Not that having an English tool-tip in a French (or whatever other locale) IDE would be a great issue; but it has to be considered (in case tool-tips are not being used).

fantaisie-software:
No need to use tooltip, the label should be sefl explanatory

kenmo-pb:
What about this? (Description is better, but it looks like a lot of empty space in the StringGadget)

image

fantaisie-software:
Description is good, can't you put the string gadget after the description instead of below ?

kenmo-pb:
Better? 🙂 

image

I had them on separate lines, because as @tajmone said, translation texts might be wider than this English text.

fantaisie-software:
Perfect. The text can still be longer, the stringgadget won't have a ton of char in it anyway

tajmone:
> The text can still be longer

In an (hypothetical) Italian translation, the space wouldn't be sufficient probably, because Italian is more verbose than English. This might be true for other locales too.

Text length aside, there's the problem that translation won't affect the width of the stringgadget. So it might be better to shrink it right away.

>  the stringgadget won't have a ton of char in it anyway

exactly, with all possible options it shouldn't require more than 4 chars, so it can be safely shrinked to 1/3 the current size IMO.

tajmone:
I also think that the description "Characters included in word selection" might be confusing, for this option pertains only "special"/"extra" (i.e. non "alphanum") chars to be considered on top of the defaults (i.e. it doesn't override the whole criteria).

Maybe "Symbols included in word selection"? (OK; a letter is a symbol too, but possibly it would be slightly more intuitive).

kenmo-pb:
Maybe just add "Extra"... `Extra characters included in word selection`?

FYI, that StringGadget is not fixed-width, the layout system is just extending it to fill the unused width. So a bit longer label would still be OK.

tajmone:
"Extra" sounds good to me, the context should make the rest clear.

> FYI, that StringGadget is not fixed-width, the layout system is just extending it to fill the unused width.

That's cool then.

fantaisie-software:
Could you squash them for easier reviewing ?

kenmo-pb:
I pushed more changes:

- better setting description
- update word chars in DiffWindow and CodeViewer too
- removed `@` from defaults

I did not add `%` to the defaults yet... I'm still concerned, see my comment from Saturday.  
Selection isn't a big deal, but this affects Find/Replace too, and I'm worried about breaking Find/Replace in modulo calculations!

image

kenmo-pb:
> Could you squash them for easier reviewing ?

Sure, let me learn how :)

SicroAtGit:
> If you rename the variable `this` by "Whole words only", then having `%` in the word chars will _not_ rename `this` in the modulo operation.

The same is already true for multiplications:
```purebasic
this = 22
that = 5

result = this*that
Debug result ; should be 110
```
That's why some PB programmers (including myself) think it's a better programming style to enclose operators with spaces so that they lose their ambiguity.

kenmo-pb:
I agree but I think compromise decisions are up to the PB team. Based on this comment from @t-harter...

> I think a config option is a good compromise. In this case I would even suggest to go with a consistent default (i.e. add * and % but not @) which is better for new users. People who want the old behavior back can easily switch it back if they want.

...I will add `%` and push one more update to this PR.

kenmo-pb:
OK I pushed one more update. `%` is now included in the default word chars.

SicroAtGit:
> I agree but I think compromise decisions are up to the PB team.

I did not speak against the compromise decision of @t-harter, but for it.

> I will add `%`

That's what I wanted to achieve. So now it's consistent.

fantaisie-software:
Tested, works perfectly !

PR #73, opened 2020-03-08
Fix Memory Viewer 'Display Hex values' setting

SicroAtGit:
The Memory Viewer setting 'Display Hex values' was set when compiling the IDE and not at runtime when opening the Memory Viewer. As a result, changing this setting had no effect.

[PB 5.71 x64] "Display Hex values" and MemoryViewer
https://www.purebasic.fr/english/viewtopic.php?f=23&t=74272


fantaisie-software:
Thx

PR #76, opened 2020-03-13
Replace ProjectOpenMode numbers with named constants

kenmo-pb:
This is a small code-cleanup PR which replaces ProjectOpenMode number literals with named constants, in preparation for more work on the Project system.


fantaisie-software:
OK

PR #77, opened 2020-03-14
Add deactivation of the `Data` checkbox in the Memory Viewer

SicroAtGit:
Setting the checkbox has no effect if the view type is `Hex` or `String`. Therefore, the checkbox should be deactivated in these cases to avoid confusion.

https://www.purebasic.fr/english/viewtopic.php?p=550837#p550837


fantaisie-software:
Perfect, thanks

Issue #80, opened 2020-03-16
Rename Files That Have Spaces in Their Names


tajmone:
I propose to rename any files that contain a space in their name, by either removing it or substituting it with `_` or `-`. Spaces in paths can interfere with automated scripts and common terminal operations, and handling them requires extra care. As a general rule, it's better to avoid spaces in Git repositories.

This is the full list of repository files containing a space in their name, which should be renamed:

- [x] `PureBasicIDE\PureBasic IDE.pbp`
- [x] `PureBasicIDE\data\logo\Logo 512.psd`
- [x] `PureBasicIDE\data\logo\PB 3D Mac Icon.bin`
- [x] `PureBasicIDE\data\logo\PB 3D Mac Icon.icns`
- [x] `PureBasicIDE\data\logo\PB 3D Mac Icon.rsc`
- [x] `PureBasicIDE\data\logo\PB 3D Win Icon.ico`

before renaming these files:

- [x] Decide how to handle spaces in each file (i.e. whether to drop the space or replace it with `_` or `-`). Different files might be better handled differently, depending on the context.
them accordingly.
- [x] Fix reference to the renamed files in other sources:
    + [x] `BUILD.md`
    + [x] `PureBasicDebugger\Makefile`
    + [x] `PureBasicIDE\data\logo\README.md`
    + [x] `PureBasicIDE\Makefile`
    + [x] `PureBasicIDE\tools\macplist.pb`

-------------------------------------------------------------------------------

I did a quick search of the repository and couldn't find any other references to these files beside those listed above.

tajmone:
I propose renaming these files the following way:

|            path           |          name         |      rename to      |
|---------------------------|-----------------------|---------------------|
| `PureBasicIDE\`           | `PureBasic IDE.pbp`   | `PureBasicIDE.pbp`  |
| `PureBasicIDE\data\logo\` | `Logo 512.psd`        | `Logo512.psd`       |
| `PureBasicIDE\data\logo\` | `PB 3D Mac Icon.bin`  | `PB3D_MacIcon.bin`  |
| `PureBasicIDE\data\logo\` | `PB 3D Mac Icon.icns` | `PB3D_MacIcon.icns` |
| `PureBasicIDE\data\logo\` | `PB 3D Mac Icon.rsc`  | `PB3D_MacIcon.rsc`  |
| `PureBasicIDE\data\logo\` | `PB 3D Win Icon.ico`  | `PB3D_WinIcon.ico`  |

[...]

Issue #24, opened 2019-12-14
Achieving Full Git Compliance


tajmone:
In its current stage, this repository suffers from usage vs Git tracking problems — i.e. in order to use and build the IDE end users inevitably change parts of its contents, thus polluting the Git working directory and ultimately having to work in branches.

This is the invevitable consequence of opening to pubblic coolaboration a project that was (until now) handled by a small number of mainteners focusing on local building. Since fixing all these issues might involve multiple steps, I've created a Project for this task, which would allow better tracking of it ongoing status:

- __[Full Git Compliance]__ (&nbsp;_project_&nbsp;) — Make repo usable without affecting tracked contents.

Furthermore, I've spotted two different categories of problems affecting the integrity of the repository:

- _machine-specific settings_ — end users need to edit tracked configuration files in order to build the IDE.
- _PureBasic IDE inteference_ — (see #15 for more details and solutions):
    + some IDE configurations on the user machine could cause source files to be modified by injecting IDE settings at the end of the source file.
    + PB project files are modified when exiting the IDE, by embedding local info which is machine-specific.

For the above reasons, I've created two separate milestones for tracking these two independent aspects of general problem covered in the __[Full Git Compliance]__ project:

- __[machine agnosis]__ (&nbsp;_milestone_&nbsp;) — Strip build toolchain of machine-dependent settings.
- __[PB vs Git]__ (&nbsp;_milestone_&nbsp;) — Protect tracked contents from changes by PB IDE.

This is just my initial proposal (and can be easily undone), but if everyone is OK with it we could use the Dashboard tools track the overall goal status by splitting it into invididual Issues per task, using milestones and the Project ([automated kanban with review]) to simplify tracking the whole thing via cross-refereces.

This issue could serve as the general discussion thread, to evaluate pros and cons and agree on individual steps, so we can create specific Issues to track each one of them.

### References

- __[Full Git Compliance]__ (&nbsp;_project_&nbsp;):
    + __[machine agnosis]__ (&nbsp;_milestone_&nbsp;)
    + __[PB vs Git]__ (&nbsp;_milestone_&nbsp;)

[Full Git Compliance]: https://github.com/fantaisie-software/purebasic/projects/1 "Go to the Project dashboard"
[PB vs Git]: https://github.com/fantaisie-software/purebasic/milestone/2 "Go to the milestone page"
[machine agnosis]: https://github.com/fantaisie-software/purebasic/milestone/1 "Go to the milestone page"
[...]

PR #81, opened 2020-03-20
Remove Spaces in Files Names

tajmone:
Rename all files containing a space in their name, and amend references in Make files and markdown documents. 

See Issue #80 regarding this commit.


SicroAtGit:
@tajmone your commit removes the file `PureBasicIDE/PureBasic IDE.pbp`

tajmone:
> @tajmone your commit removes the file `PureBasicIDE/PureBasic IDE.pbp`

Completely forgot about the ignore rule! 

Manually force added the file and amended the commit!

Thanks for catching that.

tajmone:
But I still think we shouldn't store PB project files in the repo, due to all the problems they cause to Git. We should enforce settings via compiler directive checks instead.

kenmo-pb:
> But I still think we shouldn't store PB project files in the repo, due to all the problems they cause to Git. We should enforce settings via compiler directive checks instead.

@tajmone I like the suggestion before (was it you?) that the PBP file should become "PureBasicIDE.sample.pbp" or similar, for the user to copy and customize to their liking.

Also I've been working on 2 things related to this:

1. I've created 6 pre-defined targets in the PBP file (3 OS x 2 processors) with the proper extensions (exe, ico, icns) so you can simply choose your desired target (also Build All works!) The branch is already pushed here: https://github.com/fantaisie-software/purebasic/compare/master...kenmo-pb:cross-platform-targets
2. I'm implementing a "Source control mode" setting for projects, which prevents writing all the metadata that messes with Git (last open user, file fingerprints, computer name, view settings, etc.).

tajmone:
@kenmo-pb , both features look promising. Can't wait to see them in place.

> @tajmone I like the suggestion before (was it you?) that the PBP file should become "PureBasicIDE.sample.pbp" or similar, for the user to copy and customize to their liking.

The problem there is deciding what to strip off from the proj. file (i.e. local info) and leave only the essential. I hadn't had time to dig into the `.pbp` file format and all its possible combinations (which is undocumented and need a trial and error approach to be "decoded", which is rather lengthy).

Ideally, stripping away non relevant data should be done via a script or PB tool.

fantaisie-software:
I agree about stripping all user info off the .pbp (including MD5 hash check) and create a new file with those infos. That way we can easily share the .pbp project file on a git like repo

kenmo-pb:
> I agree about stripping all user info off the .pbp (including MD5 hash check) and create a new file with those infos. That way we can easily share the .pbp project file on a git like repo

@fantaisie-software @t-harter This can be a later discussion (it won't be done for 5.72) :slightly_smiling_face: 

I have a work-in-progress but instead of saving 2 project files, it just... simply does not save that extra metadata. I have a list of what I've modified. The only functional loss so far is the "warn if files changed".

We can discuss further in issue #24 and I will push my version online soon.

tajmone:
@fantaisie-software, could you please merge this PR?

fantaisie-software:
Done

PR #83, opened 2020-03-21
Replace the window `#WINDOW_Form_Parent` with a dialog window

SicroAtGit:
By using the Dialog-Lib the window and the contained gadgets are automatically adjusted in size and position according to the contents. This avoids an oversized window with too much empty space on all
operating systems.

Improvement of pull request #50.

PB 572 b1: IDE Formdesigner display problem setting parent
https://www.purebasic.fr/english/viewtopic.php?f=4&t=74526

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -80,7 +80,7 @@ EndEnumeration
 
 ;- Windows
 ;
-Enumeration 1 ; 0 is reserved
+Runtime Enumeration 1 ; 0 is reserved
   #WINDOW_Startup
   #WINDOW_Main
   #WINDOW_About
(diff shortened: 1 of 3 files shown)


t-harter:
I don't like how the enumerations are split up. Just define the whole enum as `Runtime` as this will probably be used more and more in the future anyway.

SicroAtGit:
@t-harter Yes, I agree with you. It's easier that way. I was too careful.

SicroAtGit:
Sometimes the window is not centered on the screen, but in the upper left corner. Does anyone have a solution for this?

fantaisie-software:
You should use #PB_Window_WindowCentered instead of #PB_Window_ScreenCentered in the XML as you specify a parent window.

SicroAtGit:
> You should use #PB_Window_WindowCentered instead of #PB_Window_ScreenCentered in the XML as you specify a parent window.

Yes, I changed it.

> Also, all the variable in the IDE code begins with an uppercase (xml -> Xml)

Not really. In the code file I edited, the variables are almost always lower case:
https://github.com/fantaisie-software/purebasic/blob/master/PureBasicIDE/FormDesigner/mainevents.pb

fantaisie-software:
You're right, it's specific to the formdesigner as it was integrated in the IDE after being wrote. It's all good then !

fantaisie-software:
Tested and works fine

PR #87, opened 2020-04-29
New indentation option

Shardik-PB:
This patch introduces a new option in Preferences/Editor to indent selected source code with [Strg] + [i] based on the Tab length instead of the next function key word (default):

grafik

This is the current indentation behaviour:

This would be the indentation with the new option "Indent with Tab length" enabled and a Tab width of 2:

I had posted this request already in the PureBasic forum:
https://www.purebasic.fr/english/viewtopic.php?f=14&t=74151&start=71

Unfortunately there are also the following changes in the folder _catalogs_ necessary. The folder _catalogs_ is currently not contained in the purebasic repository!

English _catalogs/Editor.catalog_:
Insert in line 337:
`TabIndent              = Indent with Tab length`

German _catalogs/Deutsch/Editor.catalog_:
Insert in line 337:
`TabIndent              = Einrücken mit Tab-Länge`

French _catalogs/Francais/Editor.catalog_:
Insert in line 337:
`TabIndent            = Indentation avec taille des tabulation`

Spanish _catalogs/Spanish/Editor.catalog_:
Insert in line 302:
`TabIndent            = Hendidura con la anchura de la pestaña`

Italian _catalogs/Italian/Editor.catalog_:
Insert in line 321:
`TabIndent           = Incavo con Lunghezza Tabulatore`

I have translated the French, Spanish and Italian description with DeepL. It would be nice if native speakers of the respective language would check the translation.

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -300,7 +300,6 @@ Enumeration 1 ; 0 is reserved for uninitialized #PB_Any
   #GADGET_Preferences_AutoSaveAll
   #GADGET_Preferences_TabLength
   #GADGET_Preferences_RealTab
-  #GADGET_Preferences_TabIndent
   #GADGET_Preferences_MemorizeCursor
   #GADGET_Preferences_SourcePath
   #GADGET_Preferences_GetSourcePath
@@ -351,6 +350,7 @@ Enumeration 1 ; 0 is reserved for uninitialized #PB_Any
   #GADGET_Preferences_ExtraWordChars
   #GADGET_Preferences_ShowWhiteSpace
   #GADGET_Preferences_ShowIndentGuides
+  #GADGET_Preferences_UseTabIndentForSplittedLines
   #GADGET_Preferences_EnableBraceMatch
   #GADGET_Preferences_EnableKeywordMatch
   #GADGET_Preferences_SelectFont
@@ -2449,7 +2449,7 @@ Global FormVariable, FormVariableCaption, FormGrid, FormGridSize, FormEventProce
 Global FilesPanelMultiline, FilesPanelCloseButtons, FilesPanelNewButton
 Global CurrentZoom, SynchronizingZoom
 Global ExtraWordChars$
-Global TabIndent
+Global UseTabIndentForSplittedLines
 
 ; Dialog Window data
 ;
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -300,6 +300,7 @@ Enumeration 1 ; 0 is reserved for uninitialized #PB_Any
   #GADGET_Preferences_AutoSaveAll
   #GADGET_Preferences_TabLength
   #GADGET_Preferences_RealTab
+  #GADGET_Preferences_TabIndent
   #GADGET_Preferences_MemorizeCursor
   #GADGET_Preferences_SourcePath
   #GADGET_Preferences_GetSourcePath
@@ -2448,6 +2449,7 @@ Global FormVariable, FormVariableCaption, FormGrid, FormGridSize, FormEventProce
 Global FilesPanelMultiline, FilesPanelCloseButtons, FilesPanelNewButton
 Global CurrentZoom, SynchronizingZoom
 Global ExtraWordChars$
+Global TabIndent
 
 ; Dialog Window data
 ;
--- a/PureBasicIDE/Language.pb
+++ b/PureBasicIDE/Language.pb
@@ -768,6 +768,7 @@ DataSection
   Data$ "AutoSaveAll",      "Save all sources with Auto-save"
   Data$ "TabLength",        "Tab Length"
   Data$ "RealTab",          "Use real Tab (ASCII 9)"
+  Data$ "TabIndent",        "Indent with Tab Length"
   Data$ "SourcePath",       "Source Directory"
   Data$ "MemorizeCursor",   "Memorize Cursor position"
   Data$ "MemorizeMarkers",  "Memorize Marker positions"
--- a/PureBasicIDE/Language.pb
+++ b/PureBasicIDE/Language.pb
@@ -768,7 +768,7 @@ DataSection
   Data$ "AutoSaveAll",      "Save all sources with Auto-save"
   Data$ "TabLength",        "Tab Length"
   Data$ "RealTab",          "Use real Tab (ASCII 9)"
-  Data$ "TabIndent",        "Indent with Tab Length"
+  Data$ "TabIndent",        "Indent with Tab length"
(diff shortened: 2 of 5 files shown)


fantaisie-software:
It looks great, but the checkbox should be in the 'Indentation' tab, right ?

Shardik-PB:
> It looks great, but the checkbox should be in the 'Indentation' tab, right ?

In principal you are right. But the help in the tab "Preferences > Editor > Indentation" states:
> Here you can specify how the editor handles code indentation when the return key is pressed.

So all options, checkboxes etc. in this tab are only affecting the action of a pressed return key. I therefore would prefer to place the new checkbox in the Editor tab beneath "Tab Length" and "Use real Tab (ASCII 9)" because the new CheckBox "Indent with Tab length" (if enabled) is directly affected by Tab length or the use of real Tabs. It's not affected at all when pressing the return key, only by selecting a block and pressing [Strg] + [i] or selecting "Edit > Format indentation".

fantaisie-software:
But why it isn't applied when using the return key as well ? It seems wierd to have a rule for a return key indent, and another when using the block indent (ctrl+I)

Shardik-PB:
Sorry, you are right. The new feature "Indentation with Tab length" is also working with "Block mode" and "Keyword sensitive" when using the return key. I am used to always work with "No indentation". A short (incorrect) test with "Block mode" didn't indent the new line. Therefore my wrong statement about the missing indentation when using the return key. I will change my pull request and add the new option "Indent with Tab length" into the "Indentation" tab while removing it from the "Editor" tab.

t-harter:
The option also needs to be worded differently. The name currently does not reflect what the option actually does. The regular indentation applied with the Tab key already uses the tab length setting, so naming this option "Indent with Tab length" is misleading because that is already happening.

What this code change actually does is disable context sensitivity for indentation of split code lines. The name should reflect that.

tajmone:
@t-harter:

> The option also needs to be worded differently. The name currently does not reflect what the option actually does.

I agree. It seems to me that this would be an alternative way to indent single lines of code that are being split over multiple lines; so the checkbox for this option should be on an independent line in the Preferences too, to avoid confusion with the definition of tabs appearances.

Also, since this option introduces an alternative line-splitting indentation method, it might need to mention the default one too — e.g. indent by aligning to last token vs using tab indentation.

What I'm left wondering is how this new feature is going to integrate with the IDE's native formatter, which currently always fixes split-lines alignments using the default method. The code formatter should also mirror this new feature accordingly. Does it?

fantaisie-software:
What should we do with this PR ? Is it just the naming which is not right ?

t-harter:
As I wrote above:
* The naming needs to be changed
* The variable/constant naming in the code should be changed as well
* The option needs to be moved to the proper preferences page. It belongs on the indent page.

Shardik-PB:
@t-harter
> The option also needs to be worded differently. The name currently does not reflect what the option actually does. The regular indentation applied with the Tab key already uses the tab length setting, so naming this option "Indent with Tab length" is misleading because that is already happening.

> What this code change actually does is disable context sensitivity for indentation of split code lines. The name should reflect that.

> The same applies to the variable naming in the code itself.

I fully agree! Now we all need to agree on a variable and description naming which is the best reflection of its purpose which unfortunately is not easy.

My new proposal for a variable name is **_UseTabsToIndentSplitLines_** (in contrast to the default keyword sensitive indentation) and for the CheckBoxGadget's text: **_Use tabs instead of keyword  sensitive indentation for split lines_**. I will be happy for any improvement proposals.

@t-harter
> - The option needs to be moved to the proper preferences page. It belongs on the indent page.

I agree and will integrate the new option in the indentation page. I will modify my PR as soon as we have agreed on the variable and description naming.

tajmone:
Maybe, instead of **_keyword sensitive indentation for split lines_** I'd use something other than _keyword_ because that alignment can occur at various tokens types, not just keywords (e.g. strings). An alternative could be **_token-alignment for split lines_**, or maybe **_semantic-alignment_** (not sure about the latter, though).

This would also be slightly shorter to display: **_Use tabs instead of tokens-alignment for split lines,_**

Shardik-PB:
@tajmone
> This would also be slightly shorter to display: **_Use tabs instead of tokens-alignment for split lines_**

I agree about your proposition being better than mine. What about the variable name **_UseTabsToIndentSplitLines_**? Any further propositions about better naming for both?

tajmone:
@Shardik-PB:

> What about the variable name **_UseTabsToIndentSplitLines_**? Any further propositions about better naming for both?

@fantaisie-software and @t-harter have a better insight than me into the variables naming conventions used in the project, for they've created it and worked on it for years. 

Possibly, in the GUI side the features need to be spelled out more clearly, for the sake of users understanding what they do, whereas code variables could be kept shorter and more consistent with the other variables (although surely it's good to keep an intuitive association between vars identifiers and their GUI descriptions, but code comments can always fill the gap in the source, if required).

fantaisie-software:
What about 'UseTabIndentForSplittedLines' and 'Use tab indent for splitted lines' as label ?

Shardik-PB:
@tajmone and @fantaisie-software: Thank you for you propositions. But I would also like to hear from @t-harter which variable name and description he would propose.

Shardik-PB:
@tajmone 

> @fantaisie-software and @t-harter have a better insight than me into the variables naming conventions used in the project, for they've created it and worked on it for years.

I fully agree. Therefore I have taken the proposition from @fantaisie-software and implemented it in my newest commit. The variable name is now changed to **_UseTabIndentForSplittedLines_** and the label is now changed to **_Use tab indent for splitted lines_**. The new CheckBox has been moved to the preferences page **_Indentation_** as proposed by @t-harter and @fantaisie-software:

IDE-PreferencesWithNewIndentTab2

Unfortunately there are also the following changes in the folder **_catalogs_** necessary. The folder **_catalogs_** is currently not contained in the purebasic repository!

English catalogs/Editor.catalog:
Insert in line 369:
TabIndent = Use tab indent for splitted lines

German catalogs/Deutsch/Editor.catalog:
Insert in line 369:
TabIndent = Verwende Tabulator zum Einrücken von mehrzeiligem Quelltext

French catalogs/Francais/Editor.catalog:
Insert in line 369:
TabIndent = Utiliser le tiret de tabulation pour les lignes fractionnées

Spanish catalogs/Spanish/Editor.catalog:
Insert in line 329:
TabIndent = Usar tabulación para las líneas divididas

Italian catalogs/Italian/Editor.catalog:
Insert in line 352:
TabIndent = Utilizzare la linguetta di rientro per le linee divise

I have translated the French, Spanish and Italian description with DeepL. Again it would be nice if native speakers of the respective language would check and correct the translation.

fantaisie-software:
Works great.

Issue #88, opened 2020-04-30
Move PB Keywords DataSection to Separate File


tajmone:
@fantaisie-software, currently the full list of PureBasic keywords in contained in the file [`PureBasicIDE/HighlightingEngine.pb`][PBKwds] (LL 1799–1977):

```purebasic
DataSection

  ;- Keywords - BASIC

  ; Note: First is the Keyword in real Case, then
  ; the corresponding end keyword (for autocomplete), then whether the tag should include a
  ; trailing space in autocomplete (if enabled)
  ; Keywords must be sorted here!

  ; Note: Keep this in sync with the above constants

  BasicKeywords:
  Data$ "Align", "", " "
  Data$ "And", "", " "
  Data$ "Array", "", " "
  Data$ "As", "", " "
  ...
```

It would be very useful if the whole DataSection with the keywords was externalized to a separate file (just like the ASM keywords, which are stored in [`PureBasicIDE/AssemblyOperandsX86.pb`][ASMKwds]) because it would allow custom tools like syntax highlighters, editors syntaxes, etc. to easily get access to the latest PB keywords by just including the DataSection file and manipulate it in order to generate the required keywords list in the format used by the target application (e.g. in YAML, JSON, etc).

This wouldn't affect the IDE code-base, which would only require an `IncludeFile` directive inside `HighlightingEngine.pb`, but for third party tools that need only the PB/ASM keywords it would make things really simpler for it would allow to include just the keywords DataSection, without any other code.

[PBKwds]: https://github.com/fantaisie-software/purebasic/blob/master/PureBasicIDE/HighlightingEngine.pb#L1797 "View source file"
[ASMKwds]: https://github.com/fantaisie-software/purebasic/blob/master/PureBasicIDE/AssemblyOperandsX86.pb "View source file"

kenmo-pb:
Sounds like a helpful idea, and should have zero side effects on the IDE. 👍

tajmone:
> Sounds like a helpful idea, and should have zero side effects on the IDE. 

You have no idea how painful it has been to maintain my PB syntax highlighters definitions in the past years, having to extract the keywords list for the Highlighter DLL (in the SDK) of each PB release by parsing the binary DLL, and then create Lua/JSON and YAML lists for the target tools.

[...]

PR #89, opened 2020-05-13
Move BASIC Keywords to Independent File

tajmone:
Take the `DataSection` with the keywords definitions from "HighlightingEngine.pb"and move it to "KeywordsData.pbi", so that third party tools can gain access to all the PureBasic/SpiderBasic keywords definitions.

See Issue #88 for more info.


fantaisie-software:
Works fine.

PR #95, opened 2020-06-24
Image Manager - small bug fix and gui improvement

TheDarkZim:
1. The file picker has now the correct pattern.
2. The Image Manager has now a title. Needs to be added in the language catalogs, i guess?
3. The size of the window is now bigger.
4. Window can be maximized.
5. The 1. column is wider so you can read more of the imagepath.
6. The last two columns are wider so the text of the buttons fit in all languages.

After:
after
after_filepicker
Before:
before
before_filepicker


fantaisie-software:
Nice, thanks.

PR #96, opened 2020-08-09
[macOS] Revert TabBarGadget tab colors

deseven:
There were [some complaints about tab colors combination](https://www.purebasic.fr/english/viewtopic.php?p=557313#p557313) and now i see that it was indeed a bad decision to enforce system color on TabBarGadget. This PR reverts altered tab colors to their original state, all other changes related to dark mode support stay the same, including TabBarGadget background.

I still think that TabBarGadget needs an extensive rework to be able to support system themes, maybe someone someday will do that or try to approach this problem differently.


fantaisie-software:
Perfect.

PR #99, opened 2020-10-03
Prevent crash while loading a project

kenmo-pb:
If you trigger CloseProject(), for example by `Ctrl+Shift+W`, while a project is loading its sources, it crashes the IDE. This is easy to test with a large project that auto-opens many files.

This patch adds an `IsProjectBusy` flag to prevent the crash (and other odd project behaviors) by disallowing `CloseProject()` or `OpenProject()` while the current project is still busy loading (or closing) its files.


fantaisie-software (approved):
Tested and works as expected. Thanks for the fix !

PR #100, opened 2020-10-03
Fix wrong "x64" reference, and some misspellings

kenmo-pb:
In `Windows-x86.cmd` a comment tells you to specify the path of an `x64` installation - I think this is supposed to be `x86`.

While submitting this tiny correction, I'm also fixing misspellings of `special` :grin:


fantaisie-software (approved):
Perfect.

PR #102, opened 2020-10-04
Pre-define 6 cross-platform project targets

kenmo-pb:
This patch leaves the "Default Target" as-is but adds **6 new project targets**, pre-defined for Windows/Linux/Mac at 32/64-bit.
The targets have correct Build dirs, executable extensions, icon formats, and compile constants. (Please verify!)

This makes it easier to get up-and-running with the repo, or avoid reconfiguring the targets whenever switching branches or updating. It also makes Build All work (though the other OSes will display as failures, unless you choose to disable those targets).

Other minor project improvements:

- add /NOEXT param when debugging the project, to avoid re-associating your PB file extensions
- remove "5.70" from the project name
- add *.patch and *.diff extensions to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ ___*.*
 
 # Generated output files
 Build/
+*.patch
+*.diff
 
 ## ==================
 ## Binary Executables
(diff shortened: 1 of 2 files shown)


fantaisie-software:
All good

PR #104, opened 2020-10-05
Add PageUp/Dn shortcuts to step through Preferences screens

kenmo-pb:
A small feature I've wanted: Add PageUp / PageDown hotkeys to flip through the Preferences screens, regardless of where the focus is.

Only tested on Windows, but it's very basic and should work on others.

It edits `Preferences.xml` so a fresh `make` is required to regenerate `ide/Preferences.pb`!

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -205,6 +205,8 @@ Runtime Enumeration 1 ; 0 is reserved for uninitialized #PB_Any
   #GADGET_Preferences_Ok
   #GADGET_Preferences_Cancel
   #GADGET_Preferences_Apply
+  #GADGET_Preferences_PageUp    ; Not actually a gadget, for now, just a shortcut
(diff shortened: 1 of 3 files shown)


fantaisie-software (approved):
Tested on Windows, Linux and OSX. Works as expected.

fantaisie-software:
Could you squash your commits into one ?

kenmo-pb:
> Could you squash your commits into one ?

Done. This was 2 commits because right after submitting I realized it needed SetActiveGadget(). So squashing is OK.

Should I squash #102 too? (That one is not a feature change, it just makes working with the Git project easier.)

I like separate commits for separate changes. But if you prefer PR's squashed into 1 commit, I will do that.

tajmone:
> I like separate commits for separate changes.

IMO, each commit should target a specific change set, so it would make easier cherry picking, selective checkout or unrolling changes. Also, it provides a cleaner history, making it easier to focus on which files are changed by each commit. So I'd keep them separate.

PR #105, opened 2020-10-15
Show current scheme in ComboBox when opening Color Prefs

kenmo-pb:
When opening the Preferences window, and populating the ComboBox of default color schemes, this PR will automatically pre-select the user's current scheme (if it's one of those default schemes!).

This is more intuitive when opening the Color preferences, indicating what scheme is currently in use.


fantaisie-software:
Looks good.

PR #106, opened 2020-10-15
Add option to auto-wrap Find at end of file

kenmo-pb:
This PR adds a new option (a checkbox in the Find dialog) to automatically wrap searches upon reaching the end/start of file when searching forward/backward - without prompting the user with a MessageRequester.

I find this very convenient for fast F3 / Shift+F3 searching! But I left the default as the old behavior.

Note: This adds a new text string to the GUI... for now I've chosen `AutoWrap at end of file`. The exact wording can be discussed.

image


fantaisie-software:
Cool new stuff !

PR #107, opened 2020-10-18
New editor shortcuts (upper/lower/invert case)

kenmo-pb:
Here are 4 new editor actions/shortcuts requested from the forums.
By default, they're unassigned, so you have to assign them to test.

These 3 modify the selected text, or the whole word under the cursor if no selection:

- convert to Upper case
- convert to Lower case
- Invert case (example on forum, Hello --> hELLO --> Hello)

Plus, this one will select the whole word under the cursor, or grow the selection to word boundaries if already partially selected.

- Select Whole Word

**Note:** If you enable Auto Case Correction (I do) it will seem like the Case shortcuts don't work on some words... that's because it sort of conflicts with the Case Correction routine.
For example, if you try to Uppercase a keyword like `ENDIF` the IDE's Case Correction will instantly re-correct it to `EndIf`. I mentioned this in the forum before.
**My solution** is a new "less aggressive Case Correction" or "limited Case Correction" setting which I have mostly implemented, but it will need to be uploaded and thoroughly tested, since it's a big change to how Case Correction is triggered.


PureBasicIDE/ShortcutManagement.pb
```diff
@@ -763,6 +767,10 @@ CompilerElse
   #SHORTCUT_AutoCompleteConfirm= #PB_Shortcut_Tab                          ; to be tested
   #SHORTCUT_AutoCompleteAbort  = #PB_Shortcut_Escape
   #SHORTCUT_AutoIndent         = #PB_Shortcut_Command | #PB_Shortcut_I
+  #SHORTCUT_UpperCase          = 0
+  #SHORTCUT_LowerCase          = 0
+  #SHORTCUT_InvertCase         = 0
```
fantaisie-software, line 772: May be a put a default shortcut for these 3 ?
CTRL+SHIFT+U (Uppercase)
CTRL+SHIFT+L (LowerCase)
CTRL+SHIFT+X (InvertCase)

fantaisie-software (commented):
Good to me

tajmone:
Thanks @kenmo-pb, this was indeed a feature which I was missing!

> Note: If you enable Auto Case Correction (I do) it will seem like the Case shortcuts don't work on some words... that's because it sort of conflicts with the Case Correction routine.

This should be the expected behavior. But it would still allow selecting a whole line, paragraph or block, and fix its casing as desired, the Auto Case Correction would (should) then only fix the language keywords, but preserve the letter casing changes in the rest of the selection — which is what one would expect in this case.

> My solution is a new "less aggressive Case Correction" or "limited Case Correction" setting which I have mostly implemented, but it will need to be uploaded and thoroughly tested, since it's a big change to how Case Correction is triggered.

How would that behave? (just curious)

It would be great to also have a Title-Casing option — as in a _grammatically correct_ title casing function, and not like most editors implement it, i.e. by simply turning only the first letter of each word to uppercase. Of course, title casing rules vary from one language to another, so I'm referring mainly to the English language here. Also, there are different title-casing rules for English, depending on the context (legal, academic, etc.) and whether it's UK or USA English. But generally speaking, the rules proposed by the _Chicago Manual of Style_ are amongst the most used ones, and serve well most contexts:

- https://capitalizemytitle.com/style/chicago#

I personally use a lot title-casing in my source comments, for sections headings, which IMO make the code more readable.

kenmo-pb:
@fantaisie-software I pushed a new update with those 3 default shortcuts. :+1: 

@tajmone My "less aggressive" Case Correction setting (work in progress) will only case-correct the *last word you just typed*. I think this covers 99% of what people really want/expect to be auto-corrected. As it currently works, it corrects:

- entire lines at at a time
- the whole source file, as soon as you load it
- when you move the cursor over some lines
- when you paste text (though `Paste As Comment` prevents it)
- when you accidentally erase a comment `;` or a string's quote (I think this is the most complained about)
- and `Ctrl+Z` will *not* undo the case corrections!

@tajmone Title Case would be easy to add. At least, the simple version which applies it to every selected word. As soon as you start adding specific language/styling rules... personally I feel like that's the breaking line between a native IDE feature and an add-on tool.

PR #109, opened 2020-10-21
Check for external tools, when opening files

HeX0R101:
Old (own) request finally fullfilled:
https://www.purebasic.fr/english/viewtopic.php?f=18&t=70641


fantaisie-software:
Perfect

PR #110, opened 2020-11-07
Fix IDE Crash on MacOS with TouchBar

mk-soft-65:
Fix IDE Crash on MacOS with TouchBar.
Stackcorruption on RemoveSource.
Move FreeGadget to the end of Procedure.

### DIFF
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -2352,7 +2352,6 @@ Procedure RemoveSource(*Source.SourceFile = 0)
   EndIf
   
   ChangeActiveSourcecode()
-  FreeEditorGadget(Gadget)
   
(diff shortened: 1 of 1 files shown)


PureBasicIDE/SourceManagement.pb
```diff
@@ -2373,6 +2372,9 @@ Procedure RemoveSource(*Source.SourceFile = 0)
   ; There is almost no flicker anymore, so it actually looks quite good.
   FlushEvents()
   
+  ; Remove old EditorGadget
```
fantaisie-software, line 2375: Could you add a comment about why it needs to be at the very end of procedure, like:
"Fix a stack corruption on OSX, needs to be after FlushEvent()"
Because it's strange the order matters here. May be there is some event in the queue which still belongs to the editor gagdet, but it looks like a PB bug if it's the case.

Issue #113, opened 2020-11-23
New 'Documentation' Folder Fails Travis Build Due to PB Sources Problems


tajmone:
I've noticed that `master` branch is no longer passing the Travis CI build, so I investigated the matter and found it's due to several problems with the newly added PB sources in the `Documentation/` directory:

1. __Spaces in filenames__ — The original `validate.sh` script didn't expect them, but I've not updated the script and will be creating a PR with the new version that can handle them — but first we need to rebase `devel` branch on `master`, because it's lagging too many commits behind (see #84).
2. __IDE Settings in sources__ — A couple of source files contained IDE settings, so the validation failed. I've also amended this and waiting to create a PR:
3. __Trailing Whitespaces__ — numerous files contain trailing whitespace which fails the validation build. But fixing them is problematic due to the last  problem in this list.
4. __Missing BOM__ — numerous sources lack a BOM; I've tried fixing them in bulk with a dedicated tool, but I get reports that many files are not in UTF-8!
3. __Wrong Encoding__ — it seems that many sources are not in UTF-8 — worst, some batch tools are seeing them as binary, so I'm not sure whether there might be some corrupted sources in the lot.

Before trying to fix the sources, so they meet the code styles enforced on the project, I wanted to make sure that some of them were not kept this way for special reasons — e.g. for building the documentation, which might expect the PB sources to be in some specific encoding for generating the CHM file, etc.

If that was the case, then we could simply exclude some folders from the test.

Otherwise, we'll have to investigate what's wrong. I've attempted to fix them many times over, with dedicated tools, and I keep getting warnings that some sources are in binary format, which means that some corrupt characters might have crept in during some past operations (either CMD, SVN or Git operations). These things do happen, especially with old code that has been dragged around, but it's better to fix these encoding issues as soon as possible.

E.g., in old projects I often encounter non-breaking spaces or a few other stray non visible characters which are left over from the DOS era, usually resulting from file redirection and piping gone wrong in the era of CodePages.

fantaisie-software:
It's very possible than most of source are not in UTF-8. Is it only the .pb files ? There isn't a lot of them, so I can do it manually if needed.

tajmone:
Yes, it's only the PB files.

> There isn't a lot of them, so I can do it manually if needed.

[...]

PR #114, opened 2020-11-24
Fix Problems with Newly Added PB Sources

tajmone:
This PR will fix the problems with the newly added PB sources in the `Documentation/` folder, and other files which are failing code styles validation, as discussed in #113.

The PR is not ready to merge yet, for the problems have only been partially solved:

- [x] Update `validate.sh` to handle filenames with spaces.
- [x] Check and fix that some PB sources don't contain invalid characters (e.g. NBSPs `0xA0`, etc.):
    + [x] NULL chars (`0x00`).
- [x] Enforce EditorConfig code style conventions:
    + [x] Remove IDE settings from PB sources.
    + [x] Remove trailing spaces, except in indentation-only lines.
    + [x] Replace Tabs indentations with spaces.
    + [x] Convert ASCII PB sources to UTF-8, and add BOM.
    + [x] Makefiles:
        * [x] Amend Makefiles code styles settings in `.editorconfig`:
            - [x] Disable `trim_trailing_whitespace`.
            - [x] Unset `indent_style` (allow any type).
            - [x] Added `charset = utf-8`, to ensure all Makefiles are encoded in UTF-8.

        * [x] Add missing final newline in some Makefiles.
- [x] Fix code styles until they pass all Travis CI validation tests of `validate.sh`:
    + [x] No IDE settings in PB sources.
    + [x] No trailing whitespace in PB sources (except WS-only lines).
    + [x] Pass EClint validation for `.editorconfig` settings.

Since there are many files involved, and overlapping problems; I'm fixing one problem at the time, so we might want to squash a couple of commits together once all is solved, before merging.


tajmone:
# Ready to Merge

@fantaisie-software, now all PB sources and other recently added files have been fixed and their code styles amended, and the repository passes again the validation tests.

Whether to merge the PR as it is, or if you prefer me to squash some commits together (e.g. all the code-styles fixes into a single commit) I'll leave it to you. If you want me to, I'll squash some commits into one.

But maybe, since there are around 200 PB sources involved, keeping each operation in a separate commit might be better, in case we need to rollback anything — otherwise we'd have to loose _all fixes_, not just a particular fix (e.g. just trailing spaces fixes, or UTF-8 conversion).

Usually, multiple separate commits are safer in the long term, although they make the repo history a bit harder to inspect. Maybe in this case it's justified keeping them separate, for safety reasons (after all, these were automated batch-fixes, as were the validation tests; none of these files was actually tested against a real build!).

Residents/MacOS/Makefile
```diff
@@ -1,7 +1,7 @@
 # Small makefile to build the residents easily
 
 all: residents
-
```
t-harter, line 4: Some versions of make require this empty line with just a Tab for such targets to work correctly. We had problems with this in the past.

t-harter:
I don't think it is a good idea to apply the "Tabs -> Spaces" conversion to makefiles. Since you MUST use real tabs in a makefile for it to function properly there will inevitably be other places in the same file where a Tab will be used and I don't think it is very productive if somebody goes around converting tabs back to spaces after every edit in a makefile.

It is also much to easy for an overzealous conversion to introduce bugs in a makefile. Case in point: See the review comment I left above. I think you broke the makefiles for some versions of make by your changes regarding empty targets.

So I think it is best to exclude makefiles from this rule and simply accept that they will contain Tab characters.

tajmone:
> I don't think it is a good idea to apply the "Tabs -> Spaces" conversion to makefiles.

But for Makefiles we are enforcing Tabs indentation, not spaces! The indentation to spaces mentioned above refers to PB source only.

Spaces can be used freely outside of indentation, but surely I've just converted all spaces matching the indentation level to Tabs, because they were many files and had to handle them via editor functionality — but I don't recall any multiple spaces being changed within contexts other than indentation.

> Some versions of make require this empty line with just a Tab for such targets to work correctly. We had problems with this in the past.

If this is case, we can explicitly exclude those files from validation, but I think that otherwise it's a good idea to keep enforcing consistent styles, to prevent noise in commits due to different editors settings which might cause changes in indentation type at each save operation, and which might slip into commits. 

Also, it's worth noting that the problem with lines containing just a Tab failing the validation is due to the "trim trailing whitespace" rule, not the "indentation style" rule. 

I have asked many times over on the EditorConfig project to provide an option to exclude from trailing spaces checks lines that contain only whitespace, because some native language's editors (like the PB IDE, among others) insert an empty indented line between two separated indented lines, and also because if there's nothing but spaces it's indentation spaces, not trailing space, since there's nothing preceding it and lines have an EOL sequence, but not a line-start sequence.
But my requests were to no avail, the feature won't be introduced.

I'm aware that EditorConfig has many practical shortcomings (e.g. lack of support for "native" EOL, doesn't allow adding a few extra spaces to Tab indentation in order to adjust alignments, etc.), and the EClint is a very buggy and no longer developed tool, but I haven't find any better alternatives.

I'm sorry if I broke something, but the fact that Travis kept reporting a failed validation would prevent knowing if future commits were OK or not. If there are some Makefiles that need to be excluded from the validation, just tell me their full paths relative to the repository root, so I can exclude them from the "trim trailing spaces" rule.

Residents/MacOS/Makefile
tajmone: In order to fix this, I need to know if this applies only to the above mentioned file, or if this was just an example.

As mentioned further down below, after amending the lost "tab lines" I'll need to change the EditorConfig settings form Makefiles by either:

- Disable "trailing whitespace" checks for specific files, or
- Just disable the "trailing whitespaces" checks for all Makefiles.

Even if we disable these checks, we'd still have the Tab-indentation validation (which seems more important), as well as the UTF-8 validation, and other small checks.

If trailing spaces creep in the Makefiles is not a huge deal, it's just noise; but if these enforced fixes could break their functionality, that would be far worste.

fantaisie-software:
Timo is right with makefiles, it probably better to not touch them at all. That said, I'm currently running the whole build script on both PB and SB, just to ensures nothing is broken. If all is green, we can push this commit.

fantaisie-software:
Some makefiles now fails (ex: Documentation\Imports\Windows\Exports\makefile where a tab instead of space before LINKER_FORMAT=X64 broke it). Could you revert the commit about makefiles and ignore them in the travis CI file ?

tajmone:
OK, I've now reverted the previous commit that changed the Makefiles, then in a new commit I've:

- Tweaked the `.editorconfig` rules for Makefiles validation:
  - Disable `trim_trailing_whitespace` — now empty lines with just indentation will be preserved (Tabs or spaces).
  - Unset `indent_style` — Tabs or spaces can be freely used to indent now..

Now EditorConfig and Travis will only be validating Makefiles against the following code styles conventions:

- End of lines must be Unix-style `LF`.
- All Makefiles must end with an empty line.

I only fixed a few Makefiles by adding the missing final newline — this really unlikely to create problems, and it's good practice to have such an empty final line.

fantaisie-software:
Just tested it again and it works great. Thanks for the cleanup !

PR #116, opened 2020-11-25
DocMaker Help in AsciiDoc

tajmone:
# DocMaker Help in AsciiDoc

I've now also converted the DocMaker Help, taken from the `SDK\DocMaker\Readme.txt` document, and renamed it as `DocMaker-Help.asciidoc`.

An HTML converted version of the document was also added, for the benefit end users:

- [`DocMaker-Help.html`](http://htmlpreview.github.io/?https://github.com/fantaisie-software/purebasic/blob/pb-docs/Documentation/DocMaker-Help.html)

I've taken the liberty to polish and improve the original document contents in various ways, to better fit the new format and because I needed to update the obsolete link to the SVN repository.

Even though that part of the SDK is not part of this repo, I thought it would be good to have an updated version of this document inside the `Documentation/` folder, since end users might need it for contributing to the project.

If you like it, you might just as well ship it with the next package, replacing the plain text version — I believe that HTML documents are the most practical solution, they can be opened in the browser without having to launch an editor.

# Docs Improvements

In a separate commit I've also slightly improved the new AsciiDoc version of the documents relating to PB Help and DocMaker — just some text polishing, a link added here and there, and the adoption of consistent AsciiDoc headers and documents versions info.

Here are the Live HTML previews links (valid for the whole duration of the PR branch, even if new commits are added):

- [`DocMaker-Help.html`][DocMaker-Help.html]
- [`DocMaker-Tags.html`][DocMaker-Tags.html]
- [`PB-Documentation-Guide.html`][PB-Documentation-Guide.html]

[DocMaker-Help.html]: http://htmlpreview.github.io/?https://github.com/fantaisie-software/purebasic/blob/pb-docs/Documentation/DocMaker-Help.html
[DocMaker-Tags.html]: http://htmlpreview.github.io/?https://github.com/fantaisie-software/purebasic/blob/pb-docs/Documentation/DocMaker-Tags.html
[PB-Documentation-Guide.html]: http://htmlpreview.github.io/?https://github.com/fantaisie-software/purebasic/blob/pb-docs/Documentation/PB-Documentation-Guide.html

### DIFF
--- a/Documentation/DocMaker-Help.asciidoc
+++ b/Documentation/DocMaker-Help.asciidoc
@@ -1,0 +1,127 @@
+= PureBasic DocMaker Help
+Fantaisie Software <support@purebasic.com>
+:revdate: 2020-11-25
+:DocMakerVer: 5.00
+:revremark: DocMaker v{DocMakerVer}
+:revnumber!:
+:lang: en
+:favicon: https://www.purebasic.com/img/favicon.png
+// TOC Settings:
+:toc: left
+:toclevels: 5
+// Sections Numbering:
+:sectnums:
+:sectnumlevels: 2
+// Misc Settings:
+:experimental:
+:icons: font
+:linkattrs:
+:reproducible:
+:sectanchors:
+
+// IMPORTANT!!!
+// *****************************************************************************
+// If you edit the contents of this document, don't forget to update also the
+// :revdate: line accordingly (3rd line), so end users will always know which
+// version of the document they're dealing with, and which one is the latest.
+// *****************************************************************************
+
+[Preface]
+== Introduction
+
+If you want to build help files following the PureBasic format, you should use the DocMaker tool as it will lift all the hard work from you.
+
+DocMaker currently supports the following output formats:
+
+
+[horizontal]
+[.red]#*RTF*#        :: Word 97+ compatible.
+[.red]#*CHM Help*#   :: for Windows.
+[.red]#*Multiview*#  :: for Amiga.
+[.red]#*plain HTML*# :: for Linux.
+
+
+To create help files for your own libraries, modules or applications, you'll have to write a template file with some special tags, which DocMaker will convert to the target output format.
+
+
+
+== Installing DocMaker
+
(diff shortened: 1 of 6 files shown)


fantaisie-software:
Nice. I just seen the build failed because I commited a new tool CollectLanguage directly on devel. May be I should also create a pull request for every new commit ?

tajmone:
> May be I should also create a pull request for every new commit ?

We don't need to go to that extent, after all it's a dev branch, the really important thing is that `master` (soon `main`) passes the build. We can always look at the Travis report to see what's causing the fail, so we can now if a PR is the culprit or not.

But we should fix the problem soon, so we don't have to sift through the logs at each PR.

tajmone:
Th culprit seem to be `/CollectLanguage/CollectLanguage.pb` which contains trailing whitespace.

Unfortunately trimming the trailing spaces in PB sources can be tricky right now because the PB IDE doesn't remove them when using the reformatting option, and other editors/IDEs will also remove indentation lines only (which wouldn't fail the test but introduce diff noise).

Right now my approach is to first remove all trailing spaces in Sublime Text, then reformat the whole source with the PB IDE (which also reintroduces the indented empty lines).

There was a discussion on the proposal of changing PB IDE's reformatting functionality so it would remove trailing spaces, but some users mentioned that they actually exploit trailing spaces in macros definitions some times, so they opposed this feature unless optional.

Maybe the quickest solution right now would be for you to amend `/CollectLanguage/CollectLanguage.pb` and force push, since Travis is reporting the error on the `devel` baseline.

tajmone:
@fantaisie-software, I've added a PR (#118) with a commit that polishes the new PB source on `devel` so the validation checks pass.

Since the original commit was various commits behind, I didn't amend anything but just fixed it on the branch HEAD. Of course, it's still possible to move the fixing commit down and squash it with the original one, but then we'd have to force push `devel` (which might be a problem if someone is working locally on a branch therefrom).

fantaisie-software:
All good

PR #134, opened 2021-02-01
Spelling fixes

kpeters58:


fantaisie-software:
Are you sure about your change in the OS X makefile ? Makefile are a bit touchy, i would prefer to not change them

kpeters58:
Hmm - I am confused. I don't recall changing a makefile - just doc files. Tried to find the spot you are alluding to, but couldn't.
Can you please point me @ it - so I can revert the 'fix' - thanks!

Residents/MacOS/Makefile
```diff
@@ -1,9 +1,9 @@
 # Small makefile to build the residents easily
 
 all: residents
-	
 
-# We need to seperate imports and constants for the build server
```
fantaisie-software, line 6: This seems to be changed

fantaisie-software (commented):
Could you revert this file ?

Residents/MacOS/Makefile
tajmone: Looks like a whitespace change — Git sees and empty line deleted and another one added; also, the same text line is seen as different. Maybe it's related to EOL differences, tabs vs spaces, or encoding? Anyhow, it seems just noise

(we did disable some EditorConfig rules for Make files in the past, due to the need of allowing both tabs and spaces, this would explain why it passes all checks but shows as changed).

kpeters58: Yes, I edit with Sublime configured to trim trailing whitespace upon save. Apparently the help contains many lines with trailing whitespace.

tajmone: I also work with Sublime Text 3. In case you haven't already installed it, I suggest to install the EditorConfig package for ST:

https://github.com/editorconfig/editorconfig-sublime

With this package, ST will abide to the project settings, i.e. overriding your custom settings. We've worked some specific rules on a per-file-type basis on this repository, in order to enforce code style consistency.

In some cases (Make files are one of them) we can't enforce strict rules due to edge-cases occurences. In fact your PR passed all the EditorConfig validations, but Makefiles are a bit more loose when it comes to whitespace checks.

tajmone: In any case, it seems that these whitespace changes in the Make file were the accidental result of Sublime Text enforcing its own style convention at save time (e.g. a "Save all files" operation while the Makefile was open in the editor).

It should be fine to just roll-back that change. In Sublime Merge it's quite easy to also control single lines or hunks, in all types of commit editing operations.

t-harter:
Any chance to get the makefile change reverted so we can merge this?

tajmone:
@kpeters58, the simplest way to revert the Makefile to its original state is to check it out from the base branch and amend your commit.

PR #142, opened 2021-02-23
Polish OSSpecificFunctions

tajmone:
- In `Documentation/OSSpecificFunctions.txt`, polish the initial comments by wrapping them at 80 to enhance their readability.
- Adjust some alignments in the definitions themselves (as seen on most definitions).

> **NOTE** — Tested building the HTML Help files after these changes, everything worked fine.

### DIFF
--- a/Documentation/OSSpecificFunctions.txt
+++ b/Documentation/OSSpecificFunctions.txt
@@ -5,12 +5,12 @@
 ;
 ; ------------------------------------------------------------------------------
 ; 
-; This file is used to automatically generate @SupportedOS tags for the help, 
+; This file is used to automatically generate @SupportedOS tags for the help,
 ; and for any other purpose where such a list might be useful.
 ; Keep the syntax strict, as it is automatically parsed.
 ;
-; This file lists ONLY commands which do not work on all OS. Any command not listed
-; here will be marked as supported by all OS.
+; This file lists ONLY commands which do not work on all OS. Any command not
+; listed here will be marked as supported by all OS.
 ;
 ; File syntax:
 ; - Empty lines and ";" lines are ignored
@@ -23,7 +23,7 @@
 ;   <functionname> : <os tags>
 ;   <functionname> : <os tags> - <comment>
 ;
-; <os tags> can be one of these separated by ",": 
+; <os tags> can be one of these separated by ",":
 ;   Windows       - Windows, all subsystems
 ;   Linux         - Linux, all subsystems
 ;   MacOS         - OSX, all subsystems, all processor types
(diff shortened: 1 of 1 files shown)


t-harter:
I'm not a fan of this fixation on 80 characters per line. On today's monitors it is just a huge waste of screen real estate. If anything it actually decreases readability by spreading every sentence across multiple lines.

This is not the 80s anymore. Terminals with limited line lengths are a thing of the past ;)

tajmone:
> This is not the 80s anymore. Terminals with limited line lengths are a thing of the past ;)

Exactly, today developers working with modern editor split the editing pane in 2 or even 4 tabs, which allows them to compare different sources at once, hence the reason that the 80 cols rule is still found in good code styles conventions. The same applies to modern shells/terminals, which also allow splitting into multiple panes.

The wider the monitor, and the higher the resolution, the better it allows viewing multiple sources next to each other. Vertical splitting tends to be the favorite choice in horizontal monitors because it allows to see more lines at once.

tajmone:
But another good argument for keeping code within 80 columns is also for diffing (e.g. in Git when resolving conflicts), which requires a side-by-side view (splitting the screen in three side-by-side panes when dealing with three-way merges). In these cases, the operations becomes clumsy if the code ha lines longer than 80 chars, due to wrapping (which is not the default in most GUI diff tools either). 

Three-way merging done via shell/terminal is even harder beyond column 80.

Naheulf:
@tajmone How do you handle changes if one of them extends a line over 80 chars ?
 - Do you keep a longer line ? I think no because in this case it dosen't make sense to limit lines to 80 if we have exceptions
 - Do you create a new line for the overflowing text ? But after few changes it will result an alternation of 80 char lines and scrap lines (for text contents like docs)
 - Do you report the extra content on the next line and do it again until there no overflow ? But this can result a lot of changes for git even if the main change is small.

tajmone:
> @tajmone How do you handle changes if one of them extends a line over 80 chars ?

It depends. The 80 chars limit should be a guideline, not a stone-written rule, so if it's a matter of only a few chars one can make an exception. I personally think that comments should always be wrapped, and that functions/procedures can also be easily split by hard breaking the line after each parameter (if the function line is heavily indented, has a long name, and parameters are long too.

I work a lot with AsciiDoc, which (unlike markdown which recommends wrapping at 80) should split text one sentence per line — that's the idiomatic way, so I stick to it. Being prose, it makes sense because text changes can be rather chaotic, and keeping one sentence per line simplifies looking at diffs (although they wrap), whereas split sentences might lead to unexpected results since diffing algorithms might not track them nicely. But overall, AsciiDoc diffs are harder to read and require constant switching from wrapping to non-wrapping, depending on what you're comparing.

The 80 columns rule is a general guideline in typography too. Novels tend to be printed with around 80 chars per line (although non-monospace fonts lead to varying chars per line), because it seems to be the right width to allow fast reading. Movie scripts enforce monospace fonts and 80 chars per line, it's an industry standard that wasn't changed by modern printers — because it makes scripts easier and faster to read. The same is true for novels drafts. The point is that in most professions where people have to read all day long they seem to prefer text (or code) to be presented as 80 chars per line, and research in fast-reading and learning techniques corroborate this (narrower  or wider text line are harder to read).

Comments inside code are prose, intended to be read, so it makes sense to wrap it as it was a movie script (80 cols) regardless of whether the code is wrapped at 80 or not. This also ensures that textual descriptions can be easily read on the terminal. But of course, there are different "schools of thought" on this among the programmers community, just like with everything else (from tabs vs spaces, to any other dividing issue).

PR #145, opened 2021-02-24
Improve CONTRIBUTING.md

### LINKED ISSUE #144: CONTRIBUTING.md : Setting-up Your Fork for Contributions don't works
### Description of the Issue
The steps to setup the fork  for contribution don't works with a new fork and git up to date.

### Steps to Reproduce the Issue

1. Fork fantaisie-software/purebasic from the GitHub website
2. open Git Bash
3. Clone the repo with `git clone https://github.com/<github_username>/purebasic.git`
(Replace `<github_username>` with your own github username.)
4. Move to the repo usin `cd purebasic`
5. Follow steps from [CONTRIBUTING.md#setting-up-your-fork-for-contributions](https://github.com/fantaisie-software/purebasic/blob/c3bca303f9e66a4574b62e775eaf1b78ba4468aa/CONTRIBUTING.md#setting-up-your-fork-for-contributions)

### Expected Behavior
Should works.

### Actual Behavior 1
The second command generate the error: `fatal: 'upstream/devel' is not a commit and a branch 'devel' cannot be created from it`

### Fix 1
Run: `git fetch upstream` after the first command.

### Actual Behavior 2
After use the fix 1 the second command generate another error: `fatal: A branch named 'devel' already exists.`

### Fix 2
On the second command replace the `-b` (lower case B) with `-B` (uppercase B)

### More info
I let more experienced users check if theses fix are good or tricky. And I let english native speaker update the CONTRIBUTING.md file.

tajmone:
Fix some problems with the Git starter instructions, as pointed out by
@Naheulf: since the default branch was switched to `devel`, some of the
suggested commands no longer work as expected. (Fixes #144)

Polish the document and improve it:

- Add direct links to labels, to facilitate discovering pending tasks in
  specific areas of interest and competence.
- Mention the newly added repository "Discussions" area.
- Update obsolete GitHub Help URLs which were being redirected to GitHub
  Docs, and convert them into reference-style links.


CONTRIBUTING.md
```diff
@@ -187,25 +198,48 @@ Additional commits might be added to `master` branch between tagged-releases, so
...
+    $ git fetch upstream
     ```
 
-    From here-on your repository fork will recognize `upstream` as a reference to the main project (as opposed to the `origin` remote, which points to your fork on GitHub).
+    Henceforth, your repository fork will recognize `upstream` as a reference to the main project (as opposed to the `origin` remote, which points to your fork on GitHub).
+    For more info on how to add remotes, see [GitHub Docs » Adding a remote][Adding a remote].
 
-2. **TRACKING `devel`** — You then also need to ensure that your local fork of the repository contains a copy of the `devel` branch (which is not created automatically when you fork the repository):
+
+4. **TRACKING `master` BRANCH** — The default branch of the PureBasic repository is set to `devel`, not `master`, because it's main goal is development.
```
Naheulf, line 233: This step seem to be optional. If that is the case you should make it explicit.

CONTRIBUTING.md
```diff
@@ -187,25 +198,48 @@ Additional commits might be added to `master` branch between tagged-releases, so
...
+    - https://github.com/fantaisie-software/purebasic
+
+    then click on the <kbd>Fork</kbd> button that appears on the right side at beginning of the page (you need to be logged-in to see this option).
+    For more info on forking, see [GitHub Docs » Fork a repo][Fork a repo].
+
+2. **CLONE YOUR FORK LOCALLY** — You then need to _clone_ your fork locally, i.e. make a copy on your machine.
+Open up the Git Bash and type:
+
+    ```bash
+    $ git clone https://github.com/<your-GitHub-account>/purebasic.git
```
Naheulf, line 215: I know `<your-GitHub-account>` is the the placeholder syntax I used on #144 however it doesn't match the one used for `your-branch-name` below at step `2. **BRANCH**`

Naheulf (changes requested):

CONTRIBUTING.md
tajmone: It says "you _might_ also want to", which implies is not mandatory :

> So, you might also want to ensure that your local fork of the repository contains a copy of the `master `branch (which is not created automatically when you fork and clone) in order to gain access to the stable status of the project:

I think that the context is clear enough, especially if you take into account what was written before about the differences between `master `and `devel`.

I would you suggest to improve this text further?

CONTRIBUTING.md
tajmone: I'm not sure what you mean, cloning is an operation applying to a repository, not to a specific branch. Your example was correct. 

Do you mean that also the other example (the one about checking out a branch) should use angle brackets to indicate it's a placeholder? If that's the case, I don't really think we need to, for they are difference cases — the former can only be replaced by the valid username, whereas in the latter the example provided would be a valid branch name (it's an example, not a placeholder).

Naheulf: > Do you mean that also the other example (the one about checking out a branch) should use angle brackets to indicate it's a placeholder? If that's the case, I don't really think we need to, for they are difference cases — the former can only be replaced by the valid username, whereas in the latter the example provided would be a valid branch name (it's an example, not a placeholder).

I said both are placeholder because they can both be replaced with many values. (Even if the user has no choice for the first) 
That why I say both must have the same format. Either both have angle brackets or neither must have them.

tajmone: > That why I say both must have the same format. Either both have angle brackets or neither must have them.

But URLs can't contain angle brackets, whereas shell commands can. It's oranges and apples IMO. In the worst case scenario the end user will end up creating a custom branch named `your-branch-name`, which is fine in terms of creating a PR. 

The real question is whether someone who can't work out the sample command is ready to embark in Git (which is notoriously unfriendly to learn), or if he/she should become more confident with concepts like URLs, CLI tools and commands, etc. The KISS rule states that we should strive to make things simple, but not too simple as to go below the threshold of minimum requirements.

But then, if the issue is felt strongly I can just enclose the second placeholder in angle brackets too, after all the same principle of extrapolation by examples should apply. I just don't see the need for a standard convention across different example types.

Naheulf (approved):
I did not see any other unexpected behaviour sice my last comment 6 months ago about that PR or the linked issue. So, I think you can merge this PR. @tajmone or @t-harter.

tajmone:
> So, I think you can merge this PR. @tajmone 

I can't merge any PR — although I have permissions to, the agreement with Fred was to use it for other tasks only.

PR #146, opened 2021-02-28
Prevent the Save Changes prompt if a source is 'New' and also empty

kenmo-pb:
**Existing behavior**: When closing a 'New' source, the Modified flag will prompt you to save changes, even if the source is totally empty. (For example, you test a couple lines of code, then delete them.) Prompting you to save an empty file seems unnecessary.

**New behavior**: If a modified source is 'New' *and also empty*, the prompt to save changes will be prevented. (In all other conditions, the behavior remains as-is.) Notepad++ and others do this.

Feature request on the forum: https://www.purebasic.fr/english/viewtopic.php?f=3&t=76834

### DIFF
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -2409,7 +2409,20 @@ Procedure CheckSourceSaved(*Source.SourceFile = 0)
     ProcedureReturn 1
   EndIf
   
+  ; Decide if a save prompt is needed
+  PromptUser = #False
   If GetSourceModified(*Source)
+    If *Source\FileName$ <> ""
+      PromptUser = #True
+    Else
+      ; Only prompt user to save a 'New' source if it's non-empty
+      If ScintillaSendMessage(*Source\EditorGadget, #SCI_GETLENGTH) > 0
+        PromptUser = #True
+      EndIf
+    EndIf
+  EndIf
+  
+  If PromptUser
     
     ; need to make it current for display of the question
     If *Source <> *ActiveSource


t-harter:
I think this logic is better placed into the [GetSourceModified()](https://github.com/fantaisie-software/purebasic/blob/c3bca303f9e66a4574b62e775eaf1b78ba4468aa/PureBasicIDE/ScintillaHighlighting.pb#L1093) procedure itself. This way other places that rely on the source status (tab header, title bar, menu states) will reflect the same change and the behavior is more predictable to the user. As it is now, the tab title displays a "*" indicating that the source must be saved but you can close it without being asked. That is not very consistent.

kenmo-pb:
@t-harter I can move the logic there. I will search where GetSourceModified() is used, to not break anything.

For what it's worth, I often use Notepad++ as a reference for IDE behaviors.  
In Notepad++, a new-but-empty file can show the modified icon in the tab bar, and still close silently. I don't think that's a bad thing, but I agree it's inconsistent.

tajmone:
> In Notepad++, a new-but-empty file can show the modified icon in the tab bar, and still close silently. I don't think that's a bad thing, but I agree it's inconsistent.

Consistency aside (which matters) I think that it's safer to always ask about unsaved files to protect end users from losing their work — and newly created and never saved file would be a great loss, since there wouldn't be any history/snapshot left behind. It's easy to forget to save various test and draft-code files, before closing the IDE, especially when working with lots of files, under pressure and late at night, so having these type of safeguards is better than not having them.

Most editors do ask for confirmation before closing a never saved file, and I'm surprised that np++ doesn't. By the way, I consider np++ a _great_ editor, although I haven't used it much after having migrated to Sublime Text, which I find more practical and faster; but np++ has better support for Arabic text (among other things).

t-harter:
Notepad++ is not the gold standard for good editor behavior (there is probably no such thing). It has its own set of idiosyncrasies that can really drive you nuts. For example: It downright refuses to let you save a new/empty file unless you add a space first and remove it again. However, via "Save All" or "Save As" it is possible. Who decided that fresh files are not worthy of being saved but with a space in it the situation is suddenly different? :) Not all of its features are worth emulating.

tajmone:
> Notepad++ is not the gold standard for good editor behavior (there is probably no such thing)

Probably there'll never be a gold standard, after all the best behavior for each editor might also depend on the overall goals of the specific editor, its features and how these all tie in together. Which begs the question whether it might be worth drafting a gold standard specific to the PB/SB IDE, as a set of contributors' guidelines to prevent breaking its consistency.

After decades of being a closed source project, now the IDE has become open source, so it might make sense to provide similar guidelines to help those who wish to dive into its code. It doesn't have to be a full fledged specifications document, I was thinking more of adding some developer notes to the repository's Wiki, which is publicly editable. Whenever an issue relating to consistent IDE behavior comes up, like with this PR, we could take it as a chance to annotate it in the Wiki — i.e. building these guidelines as they're met in real scenarios, one topic at the time, as opposed to sitting down and writing them down systematically.

During the various exchanges inside Issues and PR's review, important topics are often unearthed and explained; it would be useful (in general) if the gist of these topic was made available in the Wiki, in some organized manner that would make them more accessible to contributors (rather than having them deeply buried in some Issue/PR discussion).

kenmo-pb:
Hi @t-harter Right, there is no gold standard. NP++ has many quirks (I agree you should be able to immediately Save a blank file...) but it's the most used Scintilla-based IDE in the world, and it does a lot of things right. That's why I referenced it. But you know the PB IDE more than anyone, and I give you the final say.

I tested this logic inside `GetSourceModified()`. Before I push the PR, I should say it trades one inconsistency for another. Now you can change a New source back to Unmodified by editing it back to its initial state. Other source files don't do that. Plus: now a New tab can look totally "brand new" when it really has some Undo history of text you wrote and deleted.

Hi @tajmone Please re-read the feature request! This PR *only affects new totally-empty files*. I don't think that's a "great loss" if the user closes it :grin: Hit `Ctrl+N` and it's back.

kenmo-pb:
I moved the logic to `GetSourceModifed()`. As is typical in programming life, it quickly grew more complicated...

- `GetSourceModified()` is called every time you add characters, or delete characters, *except when you delete the last remaining character*. And that's precisely when you would expect the Modified `*` mark to disappear. So more logic will have to be modified elsewhere to make the tab bar consistent with the suppressed save prompt.
- If we're letting a New source be edited back to Unmodified state, I don't know how that has to synchronize with the per-source variables `\ScintillaModified` `\DisplayModified` and `\ModifiedSinceUpdate`... That will take more research to make sure it's not introducing problems.
- When you create a brand New source, `Save` is *disabled in the tab's right-click menu*, but enabled in the top `File` menu. (Try it in your current PB IDE!) Is that a bug to be fixed? It's exactly the kind of inconsistency @t-harter mentioned.

I pushed an update so anyone can try it. It includes a `Debug` in `GetSourceModifed()` so you can see when that's called. This commit is not final. When finished they can be squashed together.

If nobody else wants to look into these concerns, I might close this PR for now. (It can be reopened or recreated later.) It was meant to be a quick check to suppress one `MessageRequester()`, but I have other IDE features I want to prioritize.

t-harter:
I think we should be pragmatic here and not overdo/over complicate things. Your initial solution is fine with me if the more consistent solution turns out to be a bigger issue. It is not that big of a deal after all.

Lets go with the simple solution and see how people like it. We can always come back and make it more complicated later :) An imperfect solution now is better than a perfect solution that maybe never happens.

tajmone:
> Lets go with the simple solution and see how people like it. We can always come back and make it more complicated later :) An imperfect solution now is better than a perfect solution that maybe never happens.

I agree 100% — the saying, "Less is more!" often applies. 

Also, here in Italy we have a saying: "It's better an egg today than a chicken tomorrow."

kenmo-pb:
OK, I've reverted the PR to the original simpler commit.

If you think of this feature request as just "suppress one specific MessageRequester" then I like having the logic in one place right near the MessageRequester.

Agreed that we could expand it later after carefully tracking all affected variables.

kenmo-pb:
What about my 3rd bullet point above (for a `New` unmodified file: `Save` is disabled in the tab popup menu, but enabled in the main `File` menu). Should that be considered a bug to fix? We can enable it in the popup menu.

Issue #147, opened 2021-03-03
HTML-BUILD.bat : Don't work on a non-default installation


Naheulf:
As tell is the title the [/Documentation/HTML-BUILD.bat](https://github.com/fantaisie-software/purebasic/blob/devel/Documentation/HTML-BUILD.bat) script don't works with a non-default installation.

For now I see two points that don't works:
 - it only check for `SDK\DocMaker\DocMaker.exe` in the default installation path
 - the conditional pause at the end or the file is never executed (at least on my  computer)

I found fix for theses two issues. I will create a PR soon.

tajmone:
> As tell is the title the /Documentation/HTML-BUILD.bat script don't works with a non-default installation.

no, it relies on PureBasic being installed in the default Windows' installation path (`%ProgramFiles%`):

```batch
:: Check that DocMaker is available:
SET DocMaker="%ProgramFiles%\PureBasic\SDK\DocMaker\DocMaker.exe"
IF NOT EXIST %DocMaker% GOTO :DocMakerNotFound
```

what we could do here is add an extra check, e.g. if `%DocMaker%` is already defined (by the user) when the script is launched, then use the variable as it is, instead of defining it. In this case, the script will have to take the variable for good, at face value, because `DocMaker.exe` can't be testing via the command line (e.g. `DocMaker.exe -v`) since there are command line switches to query for version info, help, etc.

Alternatively, we could check for a variable pointing to a custom PureBasic installation path, like `PUREBASIC_HOME` (which is already used by the build scripts) and then probe for the existence of `\SDK\DocMaker\DocMaker.exe` relatively to it. This approach has the advantage that it integrates better with other env vars that might have been added for other scripts in the repository (e.g. the main build scripts and Makefiles) and can reduce the number of env vars end users need to define (especially if adding them permanently to their User/System settings) — `PUREBASIC_HOME` is what these scripts are already using.


> the conditional pause at the end or the file is never executed (at least on my computer)

When creating an Issue, you should _always_ mention which OS are you working under! (if I remember correctly, from other Issues, you're using Win 7)

The script is only supposed to pause if launched from File Explorer:

```batch
:EndScript
:: Only pause before quitting if the script was launched via File Explorer:
ECHO "%cmdcmdline%" | FINDSTR /IC:"%~dp0" >nul && PAUSE
```

[...]

PR #148, opened 2021-03-03
Allow HTML-BUILD.bat to work on more configs

Naheulf:
Add some ways find DocMaker.exe if PureBasic is not installed in the default
path. It only check for values created by PureBasic tools.
Test another condition to trigger the final pause.

Fix fantaisie-software/purebasic#147

Note : Need tests on others computers for the ending conditional pause. (That why I mark it as draft)


tajmone:
# Delayed Variables Expansion

You've enabled `setlocal enableDelayedExpansion` in the script, but you're still expanding variable via `%`, instead of `!`; this might not work as expected (e.g. inside functions). From [SS64]:

> When delayed expansion is in effect, variables can be immediately read using `!variable_name!` you can still read and use `%variable_name%` but that will continue to show the initial value (expanded at the beginning of the line).

It might be safer to use the `!` notation, especially in view of future changes.



[SS64]: https://ss64.com/nt/delayedexpansion.html

Issue #120, opened 2020-12-07
PB Help for "Residents" Omits Mentioning Prototypes


tajmone:
Update PureBasic Help for "Residents", mentioning Prototypes among the list of allowed items:

- [x] PB Help English:
    [`Documentation/English/Reference/residents.txt`][ResEN]
- [x] PB Help French:
    [`Documentation/French/Reference/residents.txt`][ResFR]
- [ ] PB Help German:
    [`Documentation/German/Reference/residents.txt`][ResDE]

-------------------------------------------------------------------------------

The PureBasic documentation on Residents omits Prototypes from the list of allowed items in resident files — it only mentions structures, interfaces, macros and constants:

- [Online PB Help English » Residents][HelpEN]
- [Online PB Help French » Residents][HelpFR]
- [Online PB Help German » Residents][HelpDE]

[ResEN]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/English/Reference/residents.txt "View Help source from 'devel' branch"
[ResFR]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/French/Reference/residents.txt "View Help source from 'devel' branch"
[ResDE]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/German/Reference/residents.txt "View Help source from 'devel' branch"

[HelpEN]: https://www.purebasic.com/documentation/reference/residents.html "View online PB documentation on Residents (English)"
[HelpFR]: https://www.purebasic.com/french/documentation/reference/residents.html "View online PB documentation on Residents (French)"
[HelpDE]: https://www.purebasic.com/german/documentation/reference/residents.html "View online PB documentation on Residents (German)"

Naheulf:
I want to update the French version however GitHub show me this warning :
>  We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8. 

If i select preview it show changes for all theses characters: 'à', 'é', 'ê', 'è'. Is it safe to convert texts files from ISO-8859-1to UTF-8 (in a separate PR) ? Or should I make the change in a way that don't change encoding ?

tajmone:
> I want to update the French version however GitHub show me this warning :
> 
> > We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8.

[...]

Issue #135, opened 2021-02-22
PureBasic Help: Enforce Consistent Encoding in '.txt' Sources 


tajmone:
@Naheulf discovered that some of **PB Help**'s [`Documentation/`][Documentation] sources are encoded in ISO-8859-1, whereas most sources are in UTF-8 (see #120).

We should enforce consistent encoding in all documentation sources:

- [ ] Detect sources that are not in UTF-8 encoding and convert them from their original encoding to UTF-8:
    + [ ] [`Documentation/English/`][Documentation/English]
    + [ ] [`Documentation/French/`][Documentation/French]
    + [ ] [`Documentation/German/`][Documentation/German]

Preferably, a native speaker of the language should carry out the encoding fixes for that language version, so that he/she'll be able to inspect the affected files and check that special characters are correctly preserved (e.g. accented letters, diacritics, etc., which are frequent in German and French).

-------------------------------------------------------------------------------

@fantaisie-software, can you confirm that the documentation sources' default encoding should be UTF-8?

Or was the DocMaker created at the time PB was ASCII only, and thus expects ISO-encoded sources? Since special chars that differ between UTF-8 and ISO-8859-1 don't occur frequently, they could easily slip through corrupted and go unnoticed.

[Documentation]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation "View repo folder"

[Documentation/English]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/English
[Documentation/French]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/French
[Documentation/German]: https://github.com/fantaisie-software/purebasic/tree/devel/Documentation/German

tajmone:
# UTF-8 Validation Tests

I've written a script to quickly check if the various `.txt` sources are valid UTF-8, and as it turns out, many aren't.

The test was done by attempting to convert from UTF-8 via `iconv` (see [StackOverflow](https://stackoverflow.com/questions/115210/how-to-check-whether-a-file-is-valid-utf-8/115262#115262)):

```
$ iconv -f UTF-8 your_file > /dev/null; echo $?
```

Below are the list of non-UTF-8 files, grouped by language folder. As the results show, in the `English/` subfolder most sources are valid UTF-8 (which would make sense anyway, since it's the language with less special letters in alphabet), whereas `French/` and `German/` have a high number of non-UTF-8 files.

| Help Folder   | non-UTF8 files   |
| :------------ | ---------------: |
[...]

PR #150, opened 2021-03-06
PB Help FR » Residents: Add Prototypes

Naheulf:
In PureBasic Help (French), reference/prototypes, add Prototypes to the list of allowed items in a resident file — (along with structures, interfaces, macros and constants.

See #120 for more details.

Note: This is the French version of #121

### DIFF
--- a/Documentation/French/Reference/residents.txt
+++ b/Documentation/French/Reference/residents.txt
@@ -5,7 +5,7 @@
   compilateur. Ils se trouvent dans le dossier des 'residents' dans le chemin 
   d'installation de PureBasic. Un fichier résident doit avoir l'extension '.res' et peut 
   contenir les éléments suivants:  @ReferenceLink "structures" "Structures", 
-  @ReferenceLink "interfaces" "interfaces",  @ReferenceLink "macros" "macros" et 
+  @ReferenceLink "interfaces" "interfaces", @ReferenceLink "prototypes" "prototypes", @ReferenceLink "macros" "macros" et 
   @ReferenceLink "general_rules" "constantes". Il ne peut pas contenir un code dynamique 
   ni des @ReferenceLink "procedures" "procédures".
   @LineBreak


tajmone:
It's unbelievable that ISO encoded files show up with corrupted characters in GitHub, which makes it hard to establish if the changes are correct or not, since most special characters are shown as the [replacement character] &#xfffd;, making it impossible in languages like French, German, Italian, and others, to determine if the correct accents or diacritics are in place.

This is due to Git's lack of support for file encoding settings (which is surprising, since Git was developed in a time when code pages and ISO encoding were still in use), but it doesn't explain why GitHub can't enforce some overlay to handle encoding settings via its service.

That's why in #135 I'm pushing for the adoption of UTF-8 in DocMaker sources — there are just too many complications working with ISO files today.

Good Git GUI frontends are able to correctly autodetect file encodings and show correct source changes and diffs (e.g. [Sublime Merge]), while others (including commercial tools, like [Git Tower]) don't. Overall, applying an encoding detection algorithm shouldn't be all that hard, so I wonder why GitHub diffs can't handle properly ISO files.

Anyhow, as we can see from the [diff of this PR in GitHub's WebUI], having to deal with ISO files can be time consuming, since visual validation can't be done in the browser, but requires fetching the PR locally to inspect it with the correct tools.



[replacement character]: https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character

[Sublime Merge]: https://www.sublimemerge.com/

[Git Tower]: https://www.git-tower.com/


[diff of this PR in GitHub's WebUI]: https://github.com/fantaisie-software/purebasic/pull/150/files#diff-71f99f91f0dc84e131d8828e00da5243be0876e862bbcd6d3650ef71cb634274

Naheulf:
> It's unbelievable that ISO encoded files show up with corrupted characters in GitHub, which makes it hard to establish if the changes are correct or not, since most special characters are shown as the [replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) �, making it impossible in languages like French, German, Italian, and others, to determine if the correct accents or diacritics are in place.
> 
> This is due to Git's lack of support for file encoding settings (which is surprising, since Git was developed in a time when code pages and ISO encoding were still in use), but it doesn't explain why GitHub can't enforce some overlay to handle encoding settings via its service.

It's even worse than that: they detect the encoding (see [#120](https://github.com/fantaisie-software/purebasic/issues/120#issuecomment-783539200)) but not display it...
At least for this commit the word is the same for both English and French. So the active diff content only have full ASCII chars... It will not be the same when I will check all files for accents, diacritics and grammar errors...

> 
> That's why in #135 I'm pushing for the adoption of UTF-8 in DocMaker sources — there are just too many complications working with ISO files today.

I agree. However this need changes that only @fantaisie-software can make (see same issue)...

tajmone:
> It's even worse than that: they detect the encoding (see #120) but not display it...

In #120 you mentioned:

> I want to update the French version however GitHub show me this warning :
> 
> > We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8.

I'm not sure how you got this message. I've been working on ISO-8859-1 repositories for over three years and never encountered that message once. Also, since I push via Git I don't see how GitHub could provide me with any message at all, except for common Git error exit codes.

Is that something you got while editing files directly on GitHub's WebUI?

Also, if that was true, why the diff of this PR shows all special chars broken? If it was indeed transcoded to UTF-8, it should display correctly (which it isn't).

I'm a bit confused by this... (but, hey, I did warn that working with ISO encoding is going to raise problems all the way along, across different tools).

Naheulf:
> Is that something you got while editing files directly on GitHub's WebUI?

Yes. This is exactly what I did when I first try to edit the file.

> Also, if that was true, why the diff of this PR shows all special chars broken? If it was indeed transcoded to UTF-8, it should display correctly (which it isn't).

That why I say: "It's even worse than that: ..." encoding detection works submitting changes from the Editor WebUI but not in the Diff WebUI...

(┛◉Д◉)┛︵_qnHʇᴉꓨ_

tajmone:
> Yes. This is exactly what I did when I first try to edit the file.
> [...] encoding detection works submitting changes from the Editor WebUI but not in the Diff WebUI...

This might be something worth mentioning in `CONTRIBUING.md`, because it's quite common that contributors use the browser WebUI to quickly submit PRs with small changes; so they should be made aware that ISO encoded file would be transcoded to UTF-8 when using the WebUI.

In general, submissions via the WebUI should be discouraged, because they bypass all the local repo and code styles settings, and the validation checks that are available via `validate.sh`. But I do understand that often times someone spots a minor error and just wants to give it a quick fix without having to setup the whole workflow mumbo-jumbo on their machine.

PR #160, opened 2021-03-18
Replace SaveProjectSettings "magic numbers" with named constants

kenmo-pb:
This is another simple "code readability" PR like #76.

It uses named constants for the 4 `SaveProjectSettings` values instead of "0", "1", "2", and "3" in the code.

    #SAVESETTINGS_EndOfFile  = 0
    #SAVESETTINGS_PerFileCfg = 1
    #SAVESETTINGS_CommonCfg  = 2
    #SAVESETTINGS_DoNotSave  = 3

Yes, there are some helping comments alongside the number literals (not every time!) but I think it's safer for the constants to explain themselves.


PureBasicIDE/Common.pb
```diff
@@ -1547,6 +1547,14 @@ EndEnumeration
 
 #MATCH_Exact  = #MATCH_OS|#MATCH_Version|#MATCH_Beta|#MATCH_Processor
 
+;
+;- Values for SaveProjectSettings preference
+;
+#SAVESETTINGS_EndOfFile  = 0
+#SAVESETTINGS_PerFileCfg = 1
+#SAVESETTINGS_CommonCfg  = 2
```
Naheulf, line 1555: The IDE display "A common file project.cfg for every directory". To  match the name of the previous constant you shoul name it `#SAVESETTINGS_PerFolderCfg".
I think the "project" word should be keep away since it's unrelated to PureBasic project.

tajmone: > Edit : I don't know why GitHub select 4 lines. this comment is only for line 1555.

Because diff changes come in [hunks].

[hunks]: https://www.gnu.org/software/diffutils/manual/html_node/Hunks.html

kenmo-pb:
Hi @Naheulf, good questions.

1. Yes, it could be wrapped in an `Enumeration`. Most of the other constants are — though not all, such as `#INDENT_*` or `#INDICATOR_*`. I can change it.
2. But there is a good reason to write out the numbers. Automatic `Enumeration` are fine for constants which stay totally "inside" the IDE, like `#GADGET_*` or `#MENU_*` IDs. Those values are OK to change as we add or remove constants.
But this is a setting which is written to `.prefs` file as an integer, so **these values must never change**, or else they'll break compatibility with all the prefs files in the world! I think writing out the number emphasizes to other IDE contributors that they should not be changed or reordered. (I should add a comment.)
3. **About the naming**: This PR is not for the IDE end user, purely for us IDE developers. `SaveProjectSettings` is the name of the variable being assigned (plus `SaveProjectSettings()` is the procedure that does it) so I think it's best to keep that naming here.
4. `CommonCfg` can become `PerFolderCfg` or similar, sure :slightly_smiling_face:

t-harter:
> (I should add a comment.)

I would appreciate a comment as well. When the order or specific value of constants is important it is best to explain that right there in a comment to avoid future mistakes.

kenmo-pb:
I force-pushed a new version. It adds a comment about never changing the constant values. I also renamed `CommonCfg` to `PerFolderCfg`. `PerFolder` and `PerFile` do make sense side by side.

Issue #156, opened 2021-03-10
Bugs related to "Save As"


kenmo-pb:
Here are some related bugs concerning the `Save As` feature. **I am willing to submit fixes, if we agree on the proper desired behavior!**

**Bug 1 - Any Two Files**

1. Open two source files, for example `A.pb` and `B.pb`
2. From file `A.pb`, do a `Save As` and agree to overwrite `B.pb`
3. You now have two open tabs for the same file (this should not be allowed!)
4. Furthermore, if you have `Monitor open files for changes on disk` unchecked, you can have two different contents for the same file!

**Bug 2 - Project**

1. Open a source, `A.pb`
2. Create a project and just add `A.pb` to it
3. Do a `Save As`, for example to `C.pb`
4. The file tab renames to `C.pb` and still looks and behaves like it's part of the project
5. But only `A.pb` is still associated with the project, per the `Project Files` panel and the saved XML

**Bug 3 - Combine 1 and 2!**

1. Open two sources `A.pb` and `B.pb`
2. Create a project and just add `A.pb` to it
3. Do a `Save As` over the other file `B.pb`
4. It combines Bugs 1 & 2: You have two `B.pb` tabs, one appears to be part of the project, the other doesn't
5. Switch to the tab which *does not* appear to be part of the project
6. Press the menu item or shortcut `Add File to Project` several times (the tab won't change appearance)
7. Close the project and look at the saved XML - `B.pb` is now written several duplicate times to the project file!

**Bug 1 Solutions**

 - Disallow `Save As` over a file open in the IDE
 - Or, allow it but close out the other tab automatically

**Bug 2 Solutions**

- When doing `Save As` on a project file, the project should forget the old name and add the new name
- Or, when doing `Save As` on a project file, the project keeps the old name and does not automatically add the new

**Bug 3** should be fixed by fixing 1 and 2.

tajmone:
> __Bug 1 Solutions__
> 
> - Disallow Save As over a file open in the IDE
> - Or, allow it but close out the other tab automatically

Maybe the safest solution is to disallow saving over a file open in the IDE. Consider that the open file being overwritten might also contain unsaved changes, which would then be lost (maybe the user didn't intend this to happen, he/she accidentally selected a file (or the wrong file) from the Save As file dialog). In any case, having two versions of a same file open in the IDE is going to be problematic and shouldn't happen as you said.

> __Bug 2 Solutions__
> 
[...]

PR #161, opened 2021-03-20
Multiple "Save As" bugfixes

kenmo-pb:
Fixes `Save As` bugs, see issue #156

**1. Prevent duplicate file tabs** — If you perform `Save As` (new file or existing file) *over another file open in the IDE*, it would accept it and you could create 2 or more duplicate tabs of the same file, with unsynchronized content. Now: `SaveSourceAs()` checks the filename against all open tabs. If already open, a requester gives you the option to auto-close the other tab and lose its contents, or cancel and select a different filename.

**2. Keep tab state in sync with project** — If you have a project open, and you `Save As` from a *project file* over a *non-project file* (or the reverse), the tab color and project link were not updated. Now: This fixes that. The logic is in `HandleProjectFileSaveAs()` which resolves the correct links between `ProjectFiles()` list and `*Source`.

**3. (Minor) Suppress overwrite question for case-only change** — On case-insensitive OS (like Windows), if you `Save As` a file like "abc.pb" as "ABC.pb", the IDE asks if you're sure you want to overwrite, even though it's functionally the same file. Now: This suppresses the prompt in that special case. See (new) line 2296.

**4. (Minor) Make case-only name changes possible** — On Windows, if you try to save a file like "abc.pb" as "ABC.pb", the IDE will display the new name, but Windows actually preserves the old case on disk. Easy to verify in Windows Explorer. This is not a functional problem, but deceiving. Now: This PR includes a common workaround, which is to temporarily rename the old file on disk, so the new `Save As` has the user's intended case. See `TempFile$`. (You can also *delete* the existing file before saving, but *rename* gives the ability to undo it, should `SaveSourceFile()` fail for any reason.)

:bulb: **More Info:**

All of this uses the existing `IsEqualFile()` procedure which handles case-sensitive vs case-insensitive filename comparisons. It is the same IDE logic which determines "is this file already open?" when give a filename to load.

Per discussion in #156, `Save As` **still does not change any of your associated Project Files!** You may update them manually.

Feel free to pull this branch locally and test all the cases you can think of! With/without a project open. I have been testing for days, and I am fairly confident in all this.

**My only specific remaining concern:** In the new function `HandleProjectFileSaveAs()`, you see the `*Source` must be linked/unlinked correctly to the `ProjectFiles()` list. But I'm not sure what to do (if anything) with the `Parser.ParserData` member. Should it be cleared? Copied to the new ProjectFile from the old ProjectFile? It took me a few days to study the `ProjectFiles()` and `ParserData`, I fear only Timo and PB devs *fully* understand how it should work and correctly update.

Cheers :coffee:

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -2221,7 +2221,7 @@ Structure ProjectFile Extends ProjectFileConfig
   ; Only if the file is currently loaded
   *Source.SourceFile
   
-  ; Only if the file is corrently NOT loaded
+  ; Only if the file is currently NOT loaded
   Parser.ParserData    ; parsed source data
   
   LastOpen.l     ; valid while saving/closing a project
--- a/PureBasicIDE/Language.pb
+++ b/PureBasicIDE/Language.pb
@@ -608,6 +608,8 @@ DataSection
   Data$ "FileExists",       "The file you specified already exists!"
   Data$ "OverWrite",        "Do you want to overwrite it?"
   Data$ "CreateError",      "The file cannot be created!"
+  Data$ "FileIsOpen",       "The file you specified is currently open in the IDE!"
+  Data$ "CloseOverWrite",   "Do you want to close that tab and overwrite it?"
   
   Data$ "SaveConfigError",  "Cannot save Compiler options to file"
   Data$ "Modified",         "The file '%filename%'has been modified.%newline%Do you want to save the changes?"
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -2183,6 +2183,48 @@ Procedure OpenIncludeOnDoubleClick()
   
 EndProcedure
 
+Procedure HandleProjectFileSaveAs(*Source.SourceFile)
+  If *Source <> 0 And IsProject
+    
+    *OldProjectFile.ProjectFile = *Source\ProjectFile
+    *NewProjectFile.ProjectFile = 0
+    PushListPosition(ProjectFiles())
+    ForEach ProjectFiles()
+      If IsEqualFile(ProjectFiles()\FileName$, *Source\FileName$)
+        *NewProjectFile = @ProjectFiles()
+        Break
+      EndIf
+    Next
+    PopListPosition(ProjectFiles())
+    
+    If *NewProjectFile <> *OldProjectFile
+      If *OldProjectFile = 0
+        ; Connect source to ProjectFiles() list (new filename is part of project!)
+        If *NewProjectFile\Source <> 0
+          *NewProjectFile\Source\ProjectFile = 0
+        EndIf
+        *NewProjectFile\Source = *Source
+        *Source\ProjectFile = *NewProjectFile
+      ElseIf *NewProjectFile = 0
+        ; Disconnect source from ProjectFiles() list (new filename is not part of project!)
+        *OldProjectFile\Source = 0
+        *Source\ProjectFile = 0
+      Else
+        ; Reconnect source from one ProjectFiles() item to another (both old and new filenames are in project)
+        *OldProjectFile\Source = 0
+        *NewProjectFile\Source = *Source
+        *Source\ProjectFile = *NewProjectFile
+        *NewProjectFile\Md5$ = *OldProjectFile\Md5$
+      EndIf
+    EndIf
+    
+    If *NewProjectFile <> 0
+      ; Update name in ProjectFiles() list because case may have changed ("A.pb" -> "a.pb")
+      *NewProjectFile\FileName$ = *Source\FileName$
+    EndIf
+    
+  EndIf
+EndProcedure
 
 Procedure SaveSourceAs()
   Static NewSourcePath$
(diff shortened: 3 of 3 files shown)


t-harter:
> My only specific remaining concern: In the new function HandleProjectFileSaveAs(), you see the *Source must be linked/unlinked correctly to the ProjectFiles() list. But I'm not sure what to do (if anything) with the Parser.ParserData member. Should it be cleared? Copied to the new ProjectFile from the old ProjectFile? It took me a few days to study the ProjectFiles() and ParserData, I fear only Timo and PB devs fully understand how it should work and correctly update.

In project files the parser data is only valid when the file is not open (and autocomplete scanning is enabled for it). If in doubt, this data can be cleared and recreated any time by re-scanning the file. It is not that costly.

The [UpdateProjectFile()](https://github.com/fantaisie-software/purebasic/blob/c3bca303f9e66a4574b62e775eaf1b78ba4468aa/PureBasicIDE/ProjectManagement.pb#L335) procedure should take care of all this. It may also be a good idea to call [UnlinkSourceFromProject()](https://github.com/fantaisie-software/purebasic/blob/c3bca303f9e66a4574b62e775eaf1b78ba4468aa/PureBasicIDE/ProjectManagement.pb#L271) and [LinkSourceToProject()](https://github.com/fantaisie-software/purebasic/blob/c3bca303f9e66a4574b62e775eaf1b78ba4468aa/PureBasicIDE/ProjectManagement.pb#L298) from `HandleProjectFileSaveAs()`  to handle the linking/unlinking part correctly. Maybe it would also make sense to move the `HandleProjectFileSaveAs()` implementation into `ProjectManagement.pb` as it is more closely related to these procedures.

kenmo-pb:
Thanks for the hints @t-harter! After considering your input, and reading/testing how `Unlink` and `Link` work, I don't think my new procedure is needed at all! It's just more redundant code.

I pushed another commit. Instead of `HandleProjectFileSaveAs()` now it just calls `Unlink` before and `Link` after changing the `*ActiveSource\FileName$`. This seems to work fine. The tabs stay visually correct if you `Save As` a project file or non-project file. The `ProjectFiles()` list stays correctly updated with the `*Source` pointers.

I welcome anyone else to test this. Go wild with "Save As" - over existing files, files already open, project files --> non-project files or vice versa, or over the same type. I think it's all stable now.

------

It would be nice (but complicated) to build binaries automatically when the `devel` branch is updated. I think more people would test binaries than would fetch git code and rebuild. I don't have experience in setting that up though.

PS. @t-harter What is the policy on releasing IDE binaries on GitHub? I know any modifications must also be public open source. For example, I have some new features implemented, could I push them to my fork along with Windows binaries for other people to test?

alphasnd:
No problem about pushing your own binaries, but make it clear it's not official one, because if you got infected by a malware without knowing, we can't be responsible for that. A script to build binaries automatically here would be really cool but I don't think it's realistic as we would need to have Windows VM running somewhere to do it and it will cost money.

tajmone:
> @alphasnd: I don't think it's realistic as we would need to have Windows VM running somewhere to do it and it will cost money.

You can get free Windows VM (x64 only) via Travis CI to run your builds:

https://docs.travis-ci.com/user/reference/windows

And we're already using Travis for this repository.

tajmone:
## On Including Binaries

> @kenmo-pb: It would be nice (but complicated) to build binaries automatically when the devel branch is updated

Usually pre-built binaries are provided via Zip archives with releases (which require version tags, which the repository doesn't yet have). You can add the Zip archives manually to each release, when you create it, or have them automatically generated via a CI service like Travis — but the problem is that you'd need to upload PureBasic to the CI VM in order to do so (see #3).

Including executable binaries (i.e. Windows `.exe` files) to the repository tracked files is a bad idea because the local anti-virus might arbitrarily decide to quarantine it (as a false positive) and then Git will see that as a deleted file and won't let you switch branch (and other operations) due to changes in the work area — and of course, telling Git to undo the changes will bring the file back, which will trigger the antivirus, and so on.

DLL and object files are not usually a problem, but `.exe`s are.

t-harter:
I tested this and it seems to work fine. The only thing that I could not get to work was the "Save As" to change the case of the file name. Here the original name remains (in the IDE as well as on disk). Can somebody else give this a try?

tajmone:
> @kenmo-pb: You're probably on Windows? I ran into this too. It's a behavior of the Windows Save dialog...

If it's something handled by Windows itself it should be fine. Also keep in mind that Win 10 now allows setting directory trees to be case-sensitive, so you also need to ensure that it plays well when that's the case.

I also believe that the upcoming Win 11 edition will be pushing even more for cross-platform support features, running a Linux Kernel in the background (via WSL) and Virtual Machines, so case-sensitive directories will become more common in the nearby future, including directories being automatically set so by Windows when dealing with applications that share data across different OSs (e.g. cloud storage, severs, etc.).

t-harter:
If the Windows Save dialog prevents it anyway then the file renaming code should be removed IMHO. No need to have code around that won't do anything.

kenmo-pb:
:heavy_check_mark: I pushed an update. It removes the Windows-specific case-handling code which is not necessary.

Yes, I could squash this commit into the previous ones, but if it's OK then I prefer to preserve the commits so the Windows code snippet remains in the history, should we ever need it.

(I looked into my own branches: The Windows case-handling was there for my own `Rename File` feature I added, which piggybacks off the `Save As` feature. With `Rename File` you can rename a file's case, which on Windows requires you to temporarily do an intermediate rename.)

PR #162, opened 2021-04-10
Fix the non working triggers for saving and loading pbf files

HeX0R101:
A simple one, It just fixes that here:
https://www.purebasic.fr/english/viewtopic.php?f=42&t=77027

(sorry for the IDE comments on the bottom of the file, seems I edited it in the wrong IDE)

### DIFF
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -1700,6 +1700,7 @@ Procedure LoadSourceFile(FileName$, Activate = 1)
   If LCase(GetExtensionPart(FileName$)) = "pbf"
     OpenForm(FileName$)
     RecentFiles_AddFile(FileName$, #False)
+    AddTools_Execute(#TRIGGER_SourceLoad, *ActiveSource)
     LinkSourceToProject(*ActiveSource) ; Link To project (If any)
     ProcedureReturn 1
   EndIf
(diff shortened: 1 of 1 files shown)


HeX0R101:
There are no more Tabs now, but still fails?!

tajmone:
> There are no more Tabs now, but still fails?!

The file that doesn't pass the check is:

- `PureBasicIDE/SourceManagement.pb`

(for full details, see [Travis CI report].)

It doesn't look like there are any Tabs nor trailing spaces, but the report is for commit cbc2f6fe324405499. Maybe if you squash the commits together the error will simply go away.

I'm not sure what's wrong with this file. Either line endings are not consistent (i.e. mixing `LF` with `CRLF`) or some odd space entity creped into the source.

Are you able to run `eclint check` inside `PureBasicIDE/` and paste here the detailed report? Also, which OS are you working on?





[Travis CI report]: https://app.travis-ci.com/github/fantaisie-software/purebasic/builds/235116581

tajmone:
The problem is that I can't see any commit cbc2f6fe324405499 in this PR, the only commits currently available here are:

1. fad0e732d57173462ec437fed73e16e020b28fd9
2. 4924f66d336e17a2a606122201585450d9520cc9
3. 9b9c05f02bdb1134a9ed398210c99d925f054c45

The GitHub error report states:

> It is running a build against the merge commit, after merging #162 Fix the non working triggers for saving and loading pbf files.
Any changes that have been made to the devel branch before the build ran are also included.

I think the problem is being inherited from the base branch.

@HeX0R101, try rebasing your branch on `devel`, maybe also squash some commits (at least the `removed Tabs` commit, or all if they belong to a same changeset) and then force push, this should fix the problem.

HeX0R101:
Sorry, I'm out here, I was happy that I was somehow able to use that beast called git until now.
But I have no idea what you want me to do, too much hassle for such a Micky Mouse Fix anyway, please disregard that whole thing.

PR #163, opened 2021-05-07
Double Monitor setup and the CompilerWindow

StarBootics:
Because on Ubuntu when I have the IDE opened on Monitor #1 this tiny window appear on Monitor #2 and when the IDE is opened on Monitor#2 it appear on Monitor #2 as well but not centered. It appear in the middle of both plus an offset to the right so the entire window is on Monitor #2.

My guess is if we change on line 165 or so : 

`OpenWindow(#WINDOW_Compiler, 0, 0, 200, 50, #ProductName$, #PB_Window_ScreenCentered | #PB_Window_TitleBar | #PB_Window_Invisible, WindowID(#WINDOW_Main))`

by this line instead :  

`OpenWindow(#WINDOW_Compiler, 0, 0, 200, 50, #ProductName$, #PB_Window_WindowCentered | #PB_Window_TitleBar | #PB_Window_Invisible, WindowID(#WINDOW_Main))`

Will fix the issue on Linux operating system and will not have any impact on other operation system.


tajmone:
Something doesn't look right about this pull request, there are too many old commits showing up.

You're proposing a pull request on `master` branch, but it should be on `devel`. Can you still change the destination branch of the pull request?

StarBootics:
Apparently I can't change the branch for master. It is possible to simply delete this Request ?

tajmone:
> Apparently I can't change the branch from master to devel. 

I thought so.

> It is possible to simply delete this Request ?

Just close it and make a new one, targeting devel instead.

This happens often, it's easy to forget to check the destination branch in the pull-down menus.

SicroAtGit:
@tajmone @fantaisie-software

> This happens often, it's easy to forget to check the destination branch in the pull-down menus.

Then it would be better if the `devel` branch is the default branch.

tajmone:
> Then it would be better if the `devel` branch is the default branch.

I thought that Fred has already done this, at least I have a vague memory of this being discussed.

PR #166, opened 2021-06-21
Prevent crash when opening new project

Shardik-PB:
This patch prevents a crash on Linux when opening a new project and when SourcePath$ contains a folder path which is currently not available (for example an USB-Stick plugged out). On MacOS and Windows no crash occurs in this situation.

This problem was posted in this bug report (https://www.purebasic.fr/english/viewtopic.php?f=23&t=75479) and a workaround was described.

My added solution tests for the availability of the folder path in SourcePath$ and if not simply sets SourcePath$ to the home folder.

### DIFF
--- a/PureBasicIDE/ProjectManagement.pb
+++ b/PureBasicIDE/ProjectManagement.pb
@@ -2594,6 +2594,13 @@ Procedure OpenProjectOptions(NewProject)
       SetGadgetState(#GADGET_Project_OpenLoadLast, 1)
       
       ProjectExplorerPattern = 0 ; default pattern is "PB files"
+      CompilerIf #CompileLinux
+        ; On Linux we have to check if SourcePath$ is an accessible folder - otherwise a crash will occur!
+        ; If SourcePath$ is no folder then change it to home folder
+        If FileSize(SourcePath$) <> -2
+          SourcePath$ = "~"
+        EndIf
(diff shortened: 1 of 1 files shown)


t-harter:
I don't think automatically changing the `SourcePath$` setting is a good idea. This is a global user setting in the preferences and should not be changed without knowledge of the user (we had bug reports about such behavior in the past). I think it would be better to check for this situation and only use a different path for the actual operation that fails and not change the global setting. From a quick look it seems like doing the automatic change for the `ProjectExplorerPath$` might be the more "local" solution to this problem (if that fixes the crash).

Thinking more of it, maybe this should rather be a bugreport/bugfix done in the actual PB command that crashes with the invalid path. If this happens on only one OS then maybe making the actual PB command more resilient is the better solution also for all other PB programs. Just a thought.

Finally, I am not sure using "~" for the home folder will work correctly everywhere in the IDE. I am not sure all code can handle this path shortcut. Using `GetHomeDirectory()` may be the more safe alternative. This would also make the code cross-platform so it would not need to be Linux-specific anymore.

Shardik-PB:
@t-harter: Thank you very much for your detailed critic of my pull request. I agree with all your objections and will try to find a better solution.

t-harter:
Replaced by #177

PR #167, opened 2021-07-29
Allow Enter key to open selected Project File in list

kenmo-pb:
In the `Project` tab > `Project Files` list, you can double-click a project file to open it (or switch to it).

This PR adds the same action if you press the Enter key, useful when navigating the list by keyboard.

In fact the action is directly copied from the `#PB_EventType_LeftDoubleClick` handler in `MainWindowEvents()`.

### DIFF
--- a/PureBasicIDE/Macro.pb
+++ b/PureBasicIDE/Macro.pb
@@ -121,3 +121,9 @@ EndMacro
 Macro DebugPointer(Ptr)
   RSet(Hex(Ptr, #PB_Integer), SizeOf(INTEGER)*2, "0")
 EndMacro
+
+Macro EnsureListIconSelection(Gadget)
+  If GetGadgetState(Gadget) = -1
+    SetGadgetState(Gadget, 0)
+  EndIf
+EndMacro
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -166,7 +166,9 @@ Procedure ChangeActiveSourcecode(*OldSource.SourceFile = 0)
   ResizeMainWindow()  ; make sure the EditorGadget is correctly sized
   
   If *ActiveSource = *ProjectInfo
+    EnsureListIconSelection(#GADGET_ProjectInfo_Files)
     HideGadget(#GADGET_ProjectInfo, 0)
+    SetActiveGadget(#GADGET_ProjectInfo_Files)
     
   Else
     If *ProjectInfo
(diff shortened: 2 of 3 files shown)


t-harter:
Works ok on Windows and Mac, but doesn't work on Linux as it is.

The code you edited is only active on Windows and Mac (see the surrounding `CompilerIf`). On Linux the whole enter key handling is different (I think it is directly handled by Scintilla there or something like that) so the whole shortcut does not exist. I think a solution should be found for Linux too in order to make this a complete feature.

Two more suggestions:
* It should be enabled for the "Project Targets" list as well. There a double-click opens the compiler options so for completeness it would be good if the Enter key did the same here.
* To fully complete the ability for keyboard navigation it would be good if focus was automatically set to the ListIconGadget when you switch to the project tab. Right now, if you navigate the list and hit enter (switching tabs) and then go back to the project tab the focus will not be on the ListIconGadget and you cannot directly navigate with the keyboard anymore. Automatically setting the focus would improve that. Of course there should be some testing to ensure that this does not mess up some other workflow by taking the focus from something else.

kenmo-pb:
**Linux** - I'll look into handling the Enter key

**Project Targets List** - Agreed, for consistency. Generally I like when Double-Click and Enter do the same action in a List gadget.

**Auto-Keyboard-Focus to the List** - I like that idea. Will implement and test. Speedy keyboard navigation is nice!

kenmo-pb:
:heavy_check_mark: OK, I pushed an update:

1. Rebased onto the latest `devel`
2. Instead of duplicating the same functionality for `LeftDoubleClick` and `Return`, now the `Return` shortcut just posts a `LeftDoubleClick`
3. Project **Targets** List now reacts to the Enter key like Project Files
4. Project Files List grabs the keyboard focus when you navigate to the Project Info tab
5. You can Tab / Shift+Tab between them. Simple macro ensures a list item is selected, so it's visually obvious

:warning: Still not implemented on Linux. (There's a tricky part to every PR, huh?)

1. I don't currently have a Linux system to test on.
2. I looked into the code, tried to figure out why Linux Enter/Tab was handled differently than Windows/Mac. It was a bit more complicated than expected. There are several places in the project where Return/Tab/Shift+Tab shortcuts are dynamically added and removed (including Form Designer). Have to be careful about all cases... I considered adding the Return/Tab shortcuts when entering the Project Info tab, and removing them when leaving that tab.

This PR could either be merged without Linux functionality, or it could be shelved until me or someone else figures out the Linux handling.

t-harter:
Continued in #212.

Issue #38, opened 2019-12-30
Freeing the PureBasic IDE from all C Dependencies


tajmone:
**PROPOSAL** — Convert all C dependencies to PureBasic code and get rid of the need of C compilers in the build-chain. This is a _long term_ goal because it can't be accomplished until every C component is ported to PB.

**RATIONALE** — Having to set-up a C compiler and other C-related dependencies is a deterrent for many PB users who would otherwise like to work on the IDE project. This affects Windows users in particular, due to the need to install a full MSVS IDE (outdated version) and Win SDK, which are known to be problematic to install on modern Win versions (and this will only get worst as time passes).

Replacing the C modules with PureBasic code would also make cross-platform maintenance easier in many respects. Besides, having a native IDE written 100% in PureBasic code is a matter of pride in the language.

-------------------------------------------------------------------------------

## Looking into the C Dependencies

I've been sifting through the source code and trying to make heads and tails of what does what and where in the build chain (luckily, the comments in the sources are very useful in this respect). From what I gather, the C dependencies boil down to three distinct components:

- [ ]  [`PureBasicIDE/libmba/diff.c`][diff.c] (and its dependencies from __[libmba]__)
- [ ]  [`PureBasicIDE/AutomationDomainSocket.c`][Socket.c]
- [ ]  [`PureBasicDebugger/NetworkSupport.c`][Network.c]

The `diff.c` module from __[libmba]__ provides diffing functionality based on Eugene Myers' famous _O_(_ND_) diff algorithm (used by GNU diff), which the IDE uses to manage session history in [`PureBasicIDE/EditHistory.pb`][EditHistory.pb].

(for more info on __libmba__ and its role in this project, see [`PureBasicIDE/libmba/README.md`][libmba README]).

The other two modules deal with network functionality (especially sockets under Windows).

All of these modules should be re-implementable in PureBasic, although it might be a time-consuming task. My impression is that the diff algorithm is going to be easier part to deal with, whereas the network functions are going to be more challenging due to differences in native OS support, especially under Windows.

Also, in order to ensure that the PB-ported versions of these modules work as their C counter parts, it might be necessary to set-up some tests — e.g. a PB application that simulates their real-case usage and stress-tests both the original C modules and their PB versions, comparing the results for any discrepancies.

Although none of this is strictly necessary I think it's worth creating this Issue and putting forth the challenge.

[...]

PR #172, opened 2021-08-21
Replace libmba with a PB Diff implementation

t-harter:
Next step in removing the C dependencies from the IDE (See #38). This time replacing the diff implementation from libmba with a PB version (making the C library obsolete).

I implemented the simple Myers algorithm only (without the linear space refinement). Speed tests show that it is just as fast as the C version. I think the additional needed array space is negligible here and so the simpler algorithm is enough.

I tested with a number of files to see that a diff created with the new algorithm can be used to reconstruct the original file again (which is what we need for the history reconstruction), so I think the algorithm is correct. Still would be nice if people could test if the history feature works properly for them with these changes.


tajmone:
> @t-harter: Next step in removing the C dependencies from the IDE (See #38). This time replacing the diff implementation from libmba with a PB version (making the C library obsolete).

It's worth noting that @kenmo-pb already ported that diff algorithms to PB (see [Issue #38]), which can be seen here:


https://github.com/kenmo-pb/purebasic/blob/pb-diff-implementation/PureBasicIDE/EditHistory-Diff.pb






[Issue #38]: https://github.com/fantaisie-software/purebasic/issues/38#issuecomment-602606164

t-harter:
I saw that, but from the comments it looked like it was not ready to be merged yet. I also think that in the long run it is better to not try to mimic the libmba C interface but implement it directly in terms of the IDE data structures. That should make the code easier to maintain in the future.

Also I think the goal should be not to provide an optional replacement of the C version but get rid of it completely. This will greatly reduce the complexity of the build setup.

---

Thinking a bit more about the size requirements, I think that we do need the linear-space refinement after all. The current algorithm needs 1.5GB of memory to compare files with 10000 lines of code (not an unrealistic size). So I am converting this to a draft for now.

tajmone:
> I also think that in the long run it is better to not try to mimic the libmba C interface but implement it directly in terms of the IDE data structures. That should make the code easier to maintain in the future.

That makes sense. I haven't actually had the time to look into the PB implementation by @kenmo-pb, but as far as I can remember from various discussions it's working OK.

> Thinking a bit more about the size requirements, I think that we do need the linear-space refinement after all. The current algorithm needs 1.5GB of memory to compare files with 10000 lines of code (not an unrealistic size)

That's a lot of memory for 10.000 lines, indeed, which I guess it would trigger the paging system quite badly (possibly clogging the OS, especially under Windows). That's indeed worth looking into. Also, the original libmba algorithms are quite dated now, so chances are that the PB implementation could take advantage of modern optimizations, e.g. based on architecture and/or OS.

kenmo-pb:
Yes, all true @t-harter and @tajmone :+1: 

My version was never really production-ready, it was my first attempt at Myers diff and the code style was a little odd to drop-in mimic `libmba`. Intended for IDE contributors to avoid needing the C stuff. Mine lacks the linear space refinement, too.

Glad you're looking into this to eliminate the dependency! Maybe someone on the forums is a diff guru who could assist with the linear space version?

t-harter:
The linear space version is done. Now I am going to be dreaming about `k`, `D` and snakes for a while :smile: 

The good news is that the implementation is about 3-3.5 times faster than the original (probably because not so many callback calls are needed compared to the generic C one). Together with an improvement in the preprocessing this gives quite a speed increase.

I moved the algorithm to its own file and also reused it in the `DiffWindow.pb` code (diff tool).

kenmo-pb:
Nice job (and quickly implemented!) I will pull this into my own build of the IDE I run, for real testing.

PR #174, opened 2021-08-24
Update doc for Expat (XML internal library)

Naheulf:
Update doc on how to use Expat in recent PureBasic (> 4.50 ?) version. And add a mixed example From forum and Expat doc.


Documentation/English/Xml.txt
```diff
@@ -19,7 +19,8 @@
   be made public, the above linked license must be included with the software.
 @LineBreak
 @LineBreak  
-  The expat functions can also be called directly like other API functions with an ending underscore. This allows
+  The expat functions can also be called directly almost like other
```
Naheulf, line 22: I don't know if only Expat internal library have a "pb_" prefix or not.

Naheulf:
Notes:
- Also add a final empty line
- Still need help to update the `👄 German`  [Xml.txt](/fantaisie-software/purebasic/blob/devel/Documentation/German/Xml.txt) file.

Issue #35, opened 2019-12-27
Machine Agnostic Build Scripts


tajmone:
A major goal toward reaching full Git compliance is to ensure that end users can build the PureBasic IDE on all supported OSs without having to modify any of the repository files or affecting the Git work folder.

- [ ] Adjust repository so that the PB IDE can be compiled without affecting the Git status of the repository:
    + [ ] Provide build scripts which don't require to be edited by end users in order to build the IDE.
    + [ ] Ensure that any by-products of compiling the IDE (on all supported OSs) are ignored by Git, and that no tracked files result modified.

Currently, to build the IDE users have to modify paths in the various build scripts, which forces them to work in branches and commit their changes. This is a major factor breaking the natural Git/GitHub workflow and adds needless burdens on end users' side.

# Build Scripts Customizations

Possible solutions to the problem of end users having to customize the build scripts:

1. Provide sample build scripts (e.g. "`Window-x64.sample.cmd`") which users can copy, rename stripping the `.sample` extension suffix (e.g. "`Window-x64.cmd`") and freely edit thank to `.gitignore` rules excluding the latter from Git tracking:

        *.cmd
        !*.sample.cmd

2. Don't define inside the build scripts any environment variables which are expected to differ on each end user machine, leaving to the user the task of defining them correctly (the script should check that they are defined, and abort otherwise).
3. Avoid custom paths altogether in the build scripts. E.g. instead of letting end users provide the path of a PureBasic installation to use during the build process, assume that PureBasic was installed regularly on the machine and create a copy of the default installation into the project's build directory and use that copy.

    This approach assumes (reasonably so) that end user will be interested in building the IDE using the latest version of PureBasic, which should be already installed on the system. It also avoids under Windows any problems relating to privileges for write-accessing special OS and User folders, and problems with white spaces inside paths.
4. Find a better approach to cross-platform compilation — e.g. adding a PureBasic application able of handling all the required steps seamlessly, by auto-detecting the end user OS and available compilers, or by interactive dialogs which guide the user through the process, leaving behind a settings file ignored by Git.

Whatever the adopted solution might be, the current situation of having to edit tracked files is really problematic and poses an impediment to an effective collaborative effort.

tajmone:
## Volunteers Needed!

**This Issue needs volunteers to help with each supported OS** (not sure if this includes SpiderBasic), to guarantee a truly cross-platform solution to the problem. 

[...]

Issue #32, opened 2019-12-18
Use Windows 10 SDK


ChrisRfr:
I added Windows 10 SDK support in Window-x64(x86).cmd
They support now both Windows SDKs, following:
. Set USE_SDK10=True  (Win 10 SDK)
. Set USE_SDK10=False (Win 7.1 SDK)

In case of space in PUREBASIC_HOME path, It is possible to use the short path:
Set PUREBASIC_HOME=C:/PROGRA~1/PureBasic
In this case, we must run this batch as Administrator to have full control. I added a control on rights.

Window-x64.cmd
```
@Echo off
Cd /D %~dp0

:: Set the PureBasic home directory to a real x64 PureBasic installation
:: Warning:
::   Use a Path Without Space
::   When compiling, the IDE will be Overwriten with the new one
::   If you use C:/PROGRA~1/PureBasic short path (C:/Program Files/PureBasic). You Must run this batch as Administrator to have full control
Set PUREBASIC_HOME=C:/PROGRA~1/PureBasic
::Set PUREBASIC_HOME=E:/PureBasic/Portable/PureBasic_5.71_x64

Echo. > "%PUREBASIC_HOME%/Tmp_FullAccess"
If Exist "%PUREBASIC_HOME%/Tmp_FullAccess" (Del /q /f "%PUREBASIC_HOME%\Tmp_FullAccess") Else (Echo Run this batch as Admin to have full control on PUREBASIC_HOME path && Echo. && Pause && Goto End)

:: USE_SDK10=True for using Microsoft 10 SDK (Check SDK10_VERSION) / USE_SDK10=False for using Microsoft 7.1 SDK
Set USE_SDK10=True

Set PB_PLATEFORM_SDK=C:/Program Files/Microsoft SDKs/Windows/v7.1
Set PB_PLATEFORM_SDK10=C:/Program Files (x86)/Windows Kits/10
Set SDK10_VERSION=10.0.18362.0

Set PB_VS8=C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC
Set PATH=%PB_VS8%/bin;%PATH%
Set PATH=%PB_VS8%/bin/x86_amd64;%PATH%

:: Path to Make and dependencies from UnxUtil (cp.exe,make.exe,pwd.exe,rm.exe,touch.exe,zip.exe)
[...]

PR #178, opened 2021-10-11
Simplify build and final C cleanup

t-harter:
Now that the C dependencies are gone, we can greatly simplify the build setup because only a few preconditions remain and only a few environment variables need to be configured to run the makefiles. Most of these can actually be auto-detected. This PR removes the remaining C references and provides new and simplified build scripts:

- BuildEnv.cmd - for Windows
- BuildEnv.sh - for Linux and OSX

There is no need to edit these scripts anymore. All you need is call them with the target PB directory as an argument and the rest is configured automatically. See the content of the env scripts for details.

This PR also removes the `PB_BUILDTARGET` environment variable and stores all build files in a local `Build` directory which is not tracked by git. 

Preconditions for building are now rather simple:

- A PB installation (a separate install for IDE build and testing is recommended)
- UnixTools on Windows
- `make` and related tools on Linux/OSX

This should close the following: #38, #35, #24 (not 100% sure there)
The following issues should become obsolete by this: #32, #13

Some feedback on whether these scripts work right on various setups would be appreciated.


tajmone:
> There are still a few makefiles that do not work directly with this setup. 

You should consider switching from Make to Rake. See this article by Martin Fowler:

https://martinfowler.com/articles/rake.html

Since I've adopted Rake all my builds have become easier to setup and I've been able to handle any custom task I need.

Make is such an outdated tool, whereas Rake can do so much more than Make, especially for cross-platform project.

t-harter:
> So will it build in a local subfolder of the repo (my preference), or build and overwrite my existing PB IDE?

The IDE is still overwritten.

`PB_BUILDTARGET` used to specifiy a directory for temporary build files (compiled C sources, generated PB sources, stuff like that). The final executable was always created in the PB directory (specified via `PUREBASIC_HOME`). As such, the variable was always a bit ambiguous. I eliminated `PB_BUILDTARGET` and placed the temporary build files in a fixed subfolder of the repo. The final IDE is still copied into the PB directory. This directory has to be provided as parameter to the `BuildEnv.cmd` script.

The reason for copying the IDE into the PB directory is that you cannot test it properly otherwise. The IDE expects to be located in the PB directory and loads things like language files, themes, the compiler and other things relative to its own location. If you build it somewhere else you still have to put it somewhere resembling a PB directory to properly test it. Setting aside a dedicated PB installation to use both for compiling and as the test directory for the IDE is a simple way to solve this issue. I have worked this way on the IDE for years.

kenmo-pb:
Thanks for clarification, I understand your reasons and usage.

In my own usage, I modified the build scripts to only write locally. As long as I have copied in `Scintilla.dll` I think the IDE basically works. The main reason for my modification: **the `makefile` would give me all sorts of permission/write errors trying to write to Program Files**, depending on what PC I was working on, Administrator mode, etc. I might have to make those changes again in this new branch!

t-harter:
Why not just make a copy of your PureBasic directory somewhere else for IDE development? It does not take up that much space and you don't have to install anything. Just copy & paste your directory from Program Files to another location and you are good to go. It really can't get much simpler than that. If you start the IDE with /PORTABLE you won't even interfere with the settings of your regular install.

We could add something to .gitignore so you can copy the PB directory directly into your repository if you want to keep things in one place. This is not that much different from building locally and manually copying the needed files over to make the IDE run properly. We could even autodetect that case in `BuildEnv.cmd` so no argument will be needed.

t-harter:
The change of directory should not be enforced in the `BuildEnv.cmd` script IMHO because the build environment is for building more things in the repository than just the IDE. However, you can easily achieve this yourself by running the script from your desired location. I do it by running `BuildEnv.cmd` from a shortcut that has the current directory specified the way I want it.

I added the pause commands.

About the tools: both these commands still appear in the IDE makefile (and some other makefiles too). So better keep the old version.

ChrisRfr:
OK, Thanks for the explanations, on change directory PureBasicIDE and UnixTools for other makefiles
Just for being picky, it is preferable to use [cd /d %~dp0](https://ss64.com/nt/cd.html) rather than [cd %~dp0](https://stackoverflow.com/questions/18309941/what-does-it-mean-by-command-cd-d-dp0-in-windows/18310053) only, to change both drive and directory at once.
In case PureBasic_Opensource not in the same drive as cmd.exe (%Systemroot%/Windows32)

PR #182, opened 2021-11-05
Gtk3 support

t-harter:
Support for compilation with the Gtk3 subsystem on Linux.

This is still in the experimental phase but it already compiles and runs. If you want to give it a try and report problems that would be nice. There are probably some visual artifacts here and there due to Gtk2 and Gtk3 differences and maybe the odd crash that needs to be fixed.

In this branch Gtk3 is the default. Set `PB_GTK=2` before running `BuildEnv.sh` to switch to Gtk2.


alphasnd:
I just tried it and it does compile fine, but it doesn't run (ubuntu 17.10 x64). As soon it displays the main window, it crash with this error:

Thread 1 "purebasic" received signal SIGSEGV, Segmentation fault.
0x00007ffff2ecb8ae in XGetWindowAttributes () from /usr/lib/x86_64-linux-gnu/libX11.so.6
(gdb) bt
#0  0x00007ffff2ecb8ae in XGetWindowAttributes () at /usr/lib/x86_64-linux-gnu/libX11.so.6
#1  0x0000000000851ad7 in PB_GetWindowState ()

I will take a closer look to it to see if there is something wrong in GetWindowState()

alphasnd:
After some investigation, it appears that GetWindowState() uses X11 specifics calls, but Ubuntu 17.10 uses the Wayland window manager. So it fails to get the XWindow and crash. We will need to review all X11 specific calls in the libs and encapsulate them into something like this:

```
if (GDK_IS_WAYLAND_DISPLAY (gtk_widget_get_display(Window->Window))))
{
  // Wayland specific calls
}
else if (GDK_IS_X11_DISPLAY (gtk_widget_get_display(Window->Window))))
{
  // X11 specific calls
}

```

Once I fixed this specific crash, the IDE launch as expected :). All the windown size are too small though, but it's a very good start !

alphasnd:
PB 6.0 now works natively on Wayland, so it should be OK for GTK3. Could you merge it to devel ? I fear it will diverge too much if we wait too long

t-harter:
Should it be merged with gtk3 as default as it is in this branch or do we want to stay with gtk2 default option for a while?

Btw, can you tell what you mean with the "too small window sizes" above? I did try with Wayland myself recently (with PB6) and I noticed only some minor issues, but overall the sizes looked ok to me.

alphasnd:
I fixed the layout issue for Wayland in PB6.0 as well, so it should be good now. I would prefer still having GTK2 as default unless you're confident it will be stable enough. I plan to release SpiderBasic somewhen soon, so it will includes all the devel features.

t-harter:
I actually would prefer to do a little more work on it before merging and then make it the default (so we get more feedback). But I will be off on vacation for a while. I just did a rebase on devel so it is not diverging so far.

So I suggest to do the Spider release and merge this after that. I merged all other PRs without outstanding issues so you are good to go.

PR #183, opened 2021-11-10
Context sensitive AutoComplete for PB constants

t-harter:
In case AutoComplete is opened in a location where only a limited set of builtin PB constants make sense (for example any `Flags` parameter) the list of suggested entries is reduced to only those constants that apply in this context. This change only applies to builtin PB constants. Constants defined in the source code are still available in these locations too.

I am not sure if it is the best idea to keep the information about the constants (`ConstantsData.pbi`) within the IDE source code in the long run. It would probably be better to encode this information within the libraries themselves and read it via the compiler. That would make it easier to update this information in the future but it requires a compiler and library change.


Naheulf:
> I am not sure if it is the best idea to keep the information about the constants (`ConstantsData.pbi`) within the IDE source code in the long run.

For me this should not go in the IDE source code. However, reading it from the compiler or some external file seem good.
The last choice also allow to use it for user libraries.

ChrisRfr:
Interesting, while your hands are warm on AutoComplete, it might be nice to have the beginning bracket "**(**" already written when the word to be inserted is a function to optimize the writing by gaining a keyboard key.
ex: AutoComplete OpenWindow -> OpenWindow(

t-harter:
There is an option for this in the preferences already.

fantaisie-software:
We can put it in an external file,  may be it could be automated someday using the docmaker.

fantaisie-software:
@t-harter just tested it and seems to work fine. How did you generated the file ? I think we can merge your change and update the feature when the constants will be part of the .Desc of the libs

t-harter:
> @t-harter just tested it and seems to work fine. How did you generated the file ? I think we can merge your change and update the feature when the constants will be part of the .Desc of the libs

It is not generated: Lots of copy & paste from the help file :sweat_smile:  This is why I was thinking to change it to an external file so errors can be corrected easier.

I was thinking to wait until after the next release before merging this since we are already in the beta phase and this is a bit larger change. What do you think?

ChrisRfr:
I am not a decision maker but I think it would be good to have for the v6.00 release. This feature is really useful and an important comfort when writing code. A real bonus that would be nice to have as soon as possible.
With the big gcc changes and despite alpha and beta version, there is a strong chance of new bugs being reported after the release and with probably a 6.01 to be planned a few weeks later.
If needed, changes could be made here in that time.

fantaisie-software:
Yes, I think it's OK for 6.00, there is not much new features if we don't take in account the new C backend

fantaisie-software:
@t-harter May be we could do a specific output for the DocMaker to generate this file, what do you think ? We should have enough info if it's proper information for this (looking for all @Constant in a specific parameter. For example:

https://www.purebasic.com/documentation/window/openwindow.html

We could add #PB_Any for #Window and the all the constant for the flags. Will force us to make the doc right as well.

t-harter:
> @t-harter May be we could do a specific output for the DocMaker to generate this file, what do you think ? We should have enough info for this (looking for all @constant in a specific parameter).

I did consider that but decided against it because there were too many special cases to handle. Not all commands are documented like this and there would have to be some filtering for false-positives in place if other commands/constants are mentioned in the docs.

I think the best method in the long term is to encode this information in the library Desc somehow. We could then use it the other way around as a quality check for the docs. This seems simpler and less error prone than trying to extract the info from the Docs.

I merged the PR now as is. If we decide later against the Desc method we can still change it to an external file if we want.

PR #184, opened 2021-11-17
Enable Replace shortcut

ChrisRfr:
Enable Ctrl+H Replace shortcut


ChrisRfr:
Enable Ctrl+H Replace shortcut with the replace checkbox enabled 
If replacement is requested from a selection, use that selection as the search word and select the last entry replacement word

PureBasicIDE/FindWindow.pb
```diff
@@ -6,6 +6,7 @@
 
 
 Procedure OpenFindWindow()
+  Protected FindFromSelection.b
```
t-harter, line 9: The `.b` should be removed. Always use the integer type in normal variables even if they just store a boolean value. Using a smaller type just wastes CPU cycles converting to/from the smaller data type on every access and trying to optimize memory usage on a single variable is a waste of effort. The smaller data types should only be used in data structures.

PureBasicIDE/FindWindow.pb
```diff
@@ -32,10 +33,17 @@ Procedure OpenFindWindow()
         SetGadgetState(#GADGET_Find_SelectionOnly,FindSelectionOnly)
         SetGadgetState(#GADGET_Find_AutoWrap,     FindAutoWrap)
         
-        SetGadgetState(#GADGET_Find_DoReplace,  0) ; doreplace should always be disabled when opening this window.
-        DisableGadget(#GADGET_Find_ReplaceWord, 1)
-        DisableGadget(#GADGET_Find_Replace, 1)
-        DisableGadget( #GADGET_Find_Replaceall, 1)
+        If EventMenu() = #MENU_Replace
```
t-harter, line 36: Having the EventMenu() check here means that the whole `OpenFindWindow()` procedure can only be called in reaction to a menu event. In the interest of future flexibility this should be changed. I suggest to add a parameter for this to `OpenFindWindow()` and call it with the proper parameter depending on the menu event, just like it is done with `OpenProjectOptions()` for example.

PureBasicIDE/FindWindow.pb
ChrisRfr: > The `.b` should be removed. Always use the integer type in normal variables even if they just store a boolean value. Using a smaller type just wastes CPU cycles converting to/from the smaller data type on every access and trying to optimize memory usage on a single variable is a waste of effort. The smaller data types should only be used in data structures.

Thanks for letting me know, I didn't know. I will change it and also in other projects.
It would be good to have this, in the documentation.

PureBasicIDE/FindWindow.pb
ChrisRfr: I hesitated indeed between adding a parameter (Replace = #False) or MenuEvent(), as it is called only from UserInterface.pb
OK, I'll change it.

PureBasicIDE/ToolbarManagement.pb
```diff
@@ -562,6 +562,7 @@ DataSection
   Data.l #MENU_FindNext:            Data$ "Menu:FindNext"
   Data.l #MENU_FindPrevious:        Data$ "Menu:FindPrevious"
   Data.l #MENU_FindInFiles:         Data$ "Menu:FindInFiles"
+  Data.l #MENU_Replace:             Data$ "Menu:Replace"
```
t-harter, line 565: You need to update the count in `#NB_ToolbarMenuItems` at the top of this file for the newly added entry.

Sorry, didn't see that when I first reviewed this PR.

ChrisRfr: Well seen, my fault, I hadn't looked on how it was loaded.

t-harter: No worries. It is an easy mistake to make. Maybe this should be changed to be more like the other `DataSections` which have no predefined count but are terminated by a special value instead. Makes it easier to add new values like this.

ChrisRfr:
Oups, I was looking at the commit and I just saw an error in: Documentation/Catalogs/Deutsch/Editor.catalog 
Ersetzen          = &Ersetzen...
Should be
Replace           = &Ersetzen...
Now that it is pushed, I don't know how to fix it on my side, can you push a quick fix, please.

PR #185, opened 2021-12-01
C backend fixes

alphasnd:
- Some changes for v6.0 C compiler support


tajmone:
This PR seems to have committed an out-of-synch base branch into `devel`, dragging with it some very old unrelated commits which have been replayed on `devel` — at the risk of setting the clock backwards on the repository in terms of files changes.

alphasnd:
You're right, I messed up something.. that's why I closed the pull request. I manually merged it in devel and checked all the files, should be no problem.

tajmone:
> I manually merged it in devel and checked all the files, should be no problem.

Good. I wasn't able to check all files, since I couldn't trace the original dates of the commits, but judging from the changes to my AsciiDoc commits involved, it looked like the commits applied were later than those being replaced, which should mean that besides the commits being listed in the PR they don't carry any practical changes in the branch. 

But it's always hard to tell when this happens with many commits...

PR #187, opened 2021-12-02
MacOS GUI ToolBar position and default TabBar colors

mk-soft-65:
Modified and Ready:
- Fixed toolbar position from titlebar to top left
- Support of appearance 'Dark Mode'
- Added preferences general option 'Enable Appearance Dark Mode'
- Added preferences editor new default color schema 'Dark Mode'
- Fixed background drawing issues of ComboBox

Minimum requirement:
- Repaired StatusBar library and the background color issues or StatusBar library from PB v5.70
- macOS Build machine for Intel. I use Xcode version 12.5.1 under macOS Big Sur v11.1


mk-soft-65:
Added new preferences for enable IDE with Darkmode for macOS

mk-soft-65:
Sorry for delete and revert ...
I still don't really get along with GitHub.

PureBasicIDE/Language.pb
```diff
@@ -752,6 +752,7 @@ DataSection
   Data$ "Email",            "Email"
   Data$ "EnableMenuIcons",  "Display Icons in the Menu"
   Data$ "DisplayFullPath",  "Display full Source Path in TitleBar"
+  Data$ "DisplayDarkMode",  "Display Enable DarkMode"
```
t-harter, line 755: Not sure what "Display Enable DarkMode" is supposed to mean. "Enable Dark Mode" sounds more right to me.

PureBasicIDE/dialogs/Preferences.xml
```diff
@@ -49,6 +49,7 @@
                   <checkbox id="#GADGET_Preferences_MemorizeWindow"       lang="Preferences:MemorizeWindow" />
                   <checkbox id="#GADGET_Preferences_AutoReload"           lang="Preferences:AutoReload" />
                   <checkbox id="#GADGET_Preferences_DisplayFullPath"      lang="Preferences:DisplayFullPath" />                
+                  <checkbox id="#GADGET_Preferences_DisplayDarkMode"      lang="Preferences:DisplayDarkMode" />
```
t-harter, line 52: If the settings only has a meaning for macOS then the checkbox should be hidden (or removed) for the other OS. Otherwise we have an option which does absolutely nothing which is very frustrating for a user.

Maybe something like "Dark Mode" can be implemented for the other OS as well?

PureBasicIDE/Language.pb
mk-soft-65: Thats right. 'Enable Dark Mode' is good

PureBasicIDE/dialogs/Preferences.xml
mk-soft-65: You don't have to hide the option, but you could lock the option for other OS. Like with options that are only for Windows.

mk-soft-65:
I still have a VM with macOS High Sierra. I tested my customisation for Dark Mode and found that the TabBar colours are not correct under High Sierra. So I added the version query again. ;)

PureBasicIDE/Language.pb
mk-soft-65: Please do not forget that the PB Library 'StatusBar' from version v5.71 and higher is not yet fixed. The background colour does not work. Therefore I use the 'StatusBar' library from version v5.70 (ASM).

t-harter:
What is the reason for adding a custom `MacOSVersion()` procedure? Does the regular `OSVersion()` not work properly?

If there is a bug in the original command then it is better to file a bug report for that rather than re-implement the same functionality in the IDE code.

mk-soft-65:
The function OSVersion() returns an incorrect value '10900' under macOS v11.1. Therefore, the comparisons do not work.

Bugreport MacOS: https://www.purebasic.fr/english/viewtopic.php?t=78318

PureBasicIDE/Language.pb
mk-soft-65: Bugreport macOS StatusBar: https://www.purebasic.fr/english/viewtopic.php?t=76728

fantaisie-software:
I will add support for recent OS X version for 6.00 beta 2

mk-soft-65:
Status
- Current status taken over from fantasie-software:devel.
- Rebuild on macOS Big Sur (v11.1)
- Purebasic IDE runs on Big Sur and High Sierra ;)
- Statusbar background colour works, but when switching between aqua and darkmode only the text colour changes. With a restart of the IDE it is correct again.

mk-soft-65:
@Travis CI
why all checks have failed ?

SicroAtGit:
@mk-soft-65
```console
$ sh ./validate.sh 

===========================================
Checking PureBasic Sources for IDE Settings 
===========================================
/// Test Passed ///

==================================================
Checking PureBasic Sources for Trailing Whitespace 
==================================================
~~~ ERROR! ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The following files contain trailing whitespace:

./DialogManager/Object_Gadget.pb

Please clean up your sources and fix the problem.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/// Aborting All Tests ///

```

mk-soft-65:
I don't see problem with whitespaces !?

mk-soft-65:
After the update Beta 2 of the Purbasic Libraries the switching between Aqua and DarkMode the background colours of Windows, Splitter, Container, ToolBar does not work anymore.  

With the Beta 1 libraries this still worked!

mk-soft-65:
Fixed:

SetWindowColor(#WINDOW_Main, GetCocoaColor("windowBackgroundColor")) not longer used

mk-soft-65:
Isn't it the case that all the code from fan... devel is transferred to my devel?
So if I do a rebase, I lose my customisations and have to recreate them?

SicroAtGit:
> Isn't it the case that all the code from fan... devel is transferred to my devel?

It is the case, but because Git needs to merge two asynchronous commits histories, Git needs to create a merge commit. The pull request therefore also shows commits that actually have nothing to do with the commits in this pull request, but were from other pull requests.

> So if I do a rebase, I lose my customisations and have to recreate them?

No, Git will then update your `devel` branch to the state of the `fantaisie-software:devel` branch and the commits you had created in your `devel` will be put on top. This way no merge commit is needed, because the commit history has only been extended.

SicroAtGit:
@mk-soft-65

Forget rebasing for this PR. But @fantaisie-software should squash the commits from this PR into one commit, so that the many unnecessary merge commits don't end up in the `devel` branch.

Please use feature branches in the future instead of committing your changes directly into the `devel` branch. Otherwise it can get very ugly if you have created two pull requests and both point to your `devel` branch. If you then correct an error in one pull request with another commit in `devel`, the other pull request also gets this commit because both PR point to the `devel` branch.

t-harter (changes requested):
I would like to see these two changes before we can merge this:
* Change the language string from "Display Enable DarkMode" to "Enable Dark Mode" as discussed above
* Add code in `OpenPreferencesWindow()` to disable the gadget for this setting if the OS is not MacOS

Would also be cool if you could submit an update for the [english documentation ](https://github.com/fantaisie-software/purebasic/blob/devel/Documentation/English/Reference/ide_preferences.txt) to document this feature. The translation to other language can be handled separately, but having the english one in place would ensure that documenting the feature is not forgotten.

mk-soft-65:
But I still see a problem for the takeover.
- The error with the StatusBar library and the background colour (as reported)
- macOS Build machine for Intel. I use Xcode version 12.5.1 under macOS Big Sur v11.1 . With this version the IDE also works under High Sierra.
- When the IDE is compiled under High Sierra (Xcode 10.1), the Appearance Dark Mode does not work and ScintillaGadget has the error with the right border (as reported).

mk-soft-65:
I am so far through with it and it should work from my side ;)

fantaisie-software:
I rollback the 'fix' for the statusbar background, should be like 5.70 in 6.00 beta 3

PR #191, opened 2021-12-07
Fix the updating of the `PB_TOOL_Compiler` environment variable

SicroAtGit:
The environment variable was updated only when the compiler compiled the
source code. Therefore, only the following tool trigger events resulted
in an update of the environment variable:

- `After Compile/Run`
- `After Create Executable`

PB_TOOL_Compiler update delayed
https://www.purebasic.fr/english/viewtopic.php?t=72110

### DIFF
--- a/PureBasicIDE/AddTools.pb
+++ b/PureBasicIDE/AddTools.pb
@@ -225,7 +225,17 @@ Procedure AddTools_ExecuteCurrent(Trigger, *Target.CompileTarget)
     Protected NewList EnvVars.s()
     
     AddTools_SetEnvVar(EnvVars(), "IDE", ProgramFilename())
-    AddTools_SetEnvVar(EnvVars(), "Compiler", *CurrentCompiler\Executable$) ; use *CurrentCompiler to have the real compiler path if multiple compiler are available
+    
+    Protected *TargetCompiler.Compiler
+    If *Target\CustomCompiler
+      *TargetCompiler = FindCompiler(*Target\CompilerVersion$)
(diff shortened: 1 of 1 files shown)


SicroAtGit:
@t-harter Please merge your two commits bad0dbaf0f9c9b80ca4f65b7fa2df609d1cf3175 26e44cc21fd1d131f2c6c03b9e3126014d17ed4b quickly into the `devel` branch, because otherwise there will be more and more pull requests that fail the checks (like this one).

PureBasicIDE/AddTools.pb
```diff
@@ -225,7 +225,14 @@ Procedure AddTools_ExecuteCurrent(Trigger, *Target.CompileTarget)
     Protected NewList EnvVars.s()
     
     AddTools_SetEnvVar(EnvVars(), "IDE", ProgramFilename())
-    AddTools_SetEnvVar(EnvVars(), "Compiler", *CurrentCompiler\Executable$) ; use *CurrentCompiler to have the real compiler path if multiple compiler are available
+    
+    Protected *TargetCompiler.Compiler
+    If *Target\CustomCompiler
+      *TargetCompiler = FindCompiler(*Target\CompilerVersion$)
```
t-harter, line 231: You need to check the return value of  `FindCompiler()`. It is 0 when no matching compiler is found. So there is potential for a crash here. I suggest switching to the `DefaultCompiler` if this is the case.

SicroAtGit: I have added the change.

PR #192, opened 2021-12-29
Fix for ARM64 OS X residents

alphasnd:
- Small fix to support Apple M1


Naheulf:
```purebasic
#PB_Compiler_Processor = #PB_Processor_x64
```
is not the same as
```purebasic
#PB_Compiler_64Bit
```
?

alphasnd:
No, we have #PB_Processor_Arm64 now

PR #194, opened 2022-01-15
Fix crashing on IDE start/close, when a tool is set up to trigger in this case

SicroAtGit:
The bug was introduced by the pull request "Fix the updating of the
`PB_TOOL_Compiler` environment variable (#191)".

PB6.0 GBeta 2 Crash on start
https://www.purebasic.fr/english/viewtopic.php?t=78496

v6 Beta 2 - Invalid memory access
https://www.purebasic.fr/english/viewtopic.php?t=78503


fantaisie-software:
Thanks

PR #195, opened 2022-01-17
New tool triggers for open/file

HeX0R101:
Added three Tool Triggers to be able to distinguish between files being opened via the FileViewer and those via File/Open menu in the IDE.
See also discussion here:
https://www.purebasic.fr/english/viewtopic.php?t=77277

The problem of STARGÅTE should be also fixed now.


Naheulf:
I did not reviewed the code. However all changes in languages files must be done in all languages before merging.

HeX0R101:
Makes sense, but how am I supposed to do that?
Using deepl.com or google, which might result in (ugly) text phrases?
Or using english phrases for all languages I'm not used to, and someone else fixes that in a later PR?

SicroAtGit:
@HeX0R101 You have saved the file `German/Editor.catalog` with the line endings `CRLF` (Windows) instead of `LF` (Unix), which is why there are so many changes in this file.

HeX0R101:
No idea why that [help]-line (1022) got scrambled now, need to repush it
I'm only a little confused by the message from git:
warning: LF will be replaced by CRLF in Documentation/Catalogs/Deutsch/Editor.catalog.
What does that mean? Will I end-up at the beginning then and all LF are replaced by CRLF, again?

SicroAtGit:
> warning: LF will be replaced by CRLF in Documentation/Catalogs/Deutsch/Editor.catalog.
> What does that mean? Will I end-up at the beginning then and all LF are replaced by CRLF, again?

Git stores per default the text files internally with the line ending `LF`.
- On a Windows machine, git checks out the file then with `CRLF` line endings.
- On a Linux machine, git checks out the file then with `LF` line endings.

The warning will tell you that on next checkout the file will have the line ending `CRLF` again on your Windows machine.

To avoid this for all OS, we need to modify the `.gitattributes` file to tell git to check out this file always with the line ending `LF`. I will create a PR for this.

HeX0R101:
the non-text detection might be due to the corrupted line, no idea where that came from, but it should be fixed now.
The question about the handling of other languages is still open.

fantaisie-software:
Just put the english string if you don't know how to translate. We will take care of it when doing the doc and such

SicroAtGit:
@HeX0R101 Yes, the corrupted line was the problem. Now all looks good:
```console
$ git ls-files --eol | grep Documentation/Catalogs/Deutsch/
i/lf    w/lf    attr/text=auto        	Documentation/Catalogs/Deutsch/Compiler.catalog
i/lf    w/lf    attr/text=auto        	Documentation/Catalogs/Deutsch/Debugger.catalog
i/lf    w/lf    attr/text=auto        	Documentation/Catalogs/Deutsch/Editor.catalog
i/lf    w/lf    attr/text=auto        	Documentation/Catalogs/Deutsch/Libraries.catalog
```

@fantaisie-software Must the `*.catalog` files have a specific line ending or is it irrelevant, when it is for example `CRLF` on Windows and `LF` on Linux?

tajmone:
> the non-text detection might be due to the corrupted line, no idea where that came from, but it should be fixed now.

IRC (90% sure of it) mixed EOLs should still be detected as a text file, albeit with mixed line ending (lf/crlf), but not as binary (`-text`); so the reason might be either related to encoding (an ISO char above 127, since Git only handles UTF-8). Quite often, the culprit is a non-breaking space entity, which leads Git to think it's a binary file; these often end up in files due to text files piping in CMD using old tools, and were frequent in the old DOS (but not so much in modern Win CMD).

BTW, this used to raise a nasty bug on Git for Windows when a same file on different branches contained EOL differences only, leading to a stall due to false-positive changes that prevented switching branches — see git-for-windows/git#694. It seems that the problem has been now solved, anyhow it was possible to get out of the stall by stashing ([see my comments](https://github.com/git-for-windows/git/issues/694#issuecomment-826220195)), in case you ever find your self stuck with it.

Since Git stores EOLs info separately from file contents (internally, it only uses `LF` in the index, and normalizes to `CRLF` at checkout), these extra info (and similarly, file permissions and other attributes) can lead to problems under Windows, probably because the code that handles this stuff is still being adapted from Linux to Windows, which handle them quite differently.

HeX0R101:
Would it be possible to check this PR and maybe integrate it into PB6.0?
Because in fact it is not only an enhancement, but also a fix for my [PR #109 ](https://github.com/fantaisie-software/purebasic/pull/109)

fantaisie-software:
Just tested and looks good to me. Thanks !

PR #204, opened 2022-02-16
Fixed some issues

mk-soft-65:
AutoComplete:
- Fix RemoveKeyBoardShortcut constants #PB_Shortcut_*

macOS:
- Fix more ComboboxGadget background drawing

macOS Preference Compiler:
- Restart OpenFileRequester to open app packet folder 'Contents'

### DIFF
--- a/DialogManager/Object_Gadget.pb
+++ b/DialogManager/Object_Gadget.pb
@@ -122,12 +122,7 @@ Procedure DlgGadget_New(*StaticData.DialogObjectData)
         *THIS\HasTitle = #False
         
       Case #DIALOG_ComboBox
-        CompilerIf #CompileMac
-          ; Fixed macOS background drawing issue from the ComboBox, needs minimum size
-          *THIS\Gadget = ComboBoxGadget(*StaticData\Gadget, 0, 0, 100, 25, *StaticData\Flags)
-        CompilerElse
-          *THIS\Gadget = ComboBoxGadget(*StaticData\Gadget, 0, 0, 0, 0, *StaticData\Flags)
-        CompilerEndIf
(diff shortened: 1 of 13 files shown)


mk-soft-65:
No found whitespaces on common.pb. 
But not any changes from me.

HeX0R101:
Wouldn't it be better to let the PB team fix the ComboBox behaviour for MacOS in PB itself, instead of adding so many compiler directives to the PB IDE?

mk-soft-65:
Maybe it's better.

But I will remove the many compilerIFs and do them via a macro.

t-harter:
I am not the biggest fan of the "replace a PB command with a macro" trick. It can lead to hard to find bugs because it is simply not expected that a builtin command is not actually the original. Debugging something like that is really frustrating.

A better solution is to have a procedure named `ComboBoxGadgetSizeFix` or similar in all `XxxExtensions.pb` files which simply wraps `ComboBoxGadget()` for the OS where the fix is not needed. Then use this procedure everywhere in the code. This makes it much clearer what is going on when you read the code.

mk-soft-65:
That may be correct, 
but it is a bug from ComboGadget in the macOS Gadget library or macOS framework and not in the PB IDE code. It should probably be fixed there. This way the customisation can be removed faster afterwards.

t-harter:
It doesn't make much sense to me to have bugs in PB commands and then have workarounds for them in the official PB IDE. If there is a bug in a PB command it must be fixed there and not in the IDE at all.

Besides, who guarantees that it will be removed afterwards? If I know one thing about code maintenance it is that the bad hacks always live longer than expected and things that should just be temporary tend to be forgotten and left in the code. That goes double if it is actually invisible because it is hidden behind a macro. So you may just as well write code as if it were permanent.

mk-soft-65:
I am goback to CompilerIf's

PR #207, opened 2022-05-15
Drop PB Trailing Spaces Validation

tajmone:
Tweak the custom `validate.sh` script so the Travis CI build no longer
checks that PureBasic sources don't contain trailing spaces.

This check was performed via a custom Bash script, not via EClint,
so the `.editorconfig` settings were not affected by this commit.


fantaisie-software:
Thanks !

PR #210, opened 2022-06-12
Small fix macOS darkmode and auto complete

mk-soft-65:
Under macOS, the TabBarPanel no longer functions to switch between DarkMode and Aqua.
This is because the function 'GetTabBarGadgetItemColor' no longer returns the previously set color values (with alphachannel and signed).

The auto-completion for 'RemoveKeyboardShortcut' constants is not correct.


fantaisie-software:
Thanks !

PR #221, opened 2022-12-08
Replace All Undo/Redo

ChrisRfr:
Undo/Redo done in one Go, after a "Replace All". 
See: https://www.purebasic.fr/english/viewtopic.php?p=592319#p592319


fantaisie-software:
Tested and works as expected, thanks !

PR #225, opened 2023-03-12
fix(DebugOutput): avoid macOS IDE crash upon empty debug message

### LINKED ISSUE #224: Debugging an empty string on macOS crashes the IDE
## Platform
**PB:** 6.01 LTS  (aa1dbc85256c)
**OS:** macOS 13.2.1, arm64

## What happened?

The following simple code will crash the IDE when run with debugger active:

```pb
Debug ""
```

It doesn't matter if given a string literal or relying on a variable. This also crashes:

```pb
Define someInputValue.s = ""    ; Pretend we read no input.
Debug  someInputValue           ; Debugging this crashes us.
```

## What is the expected behaviour?

The IDE should not crash when debugging empty strings.

## Why does this happen?

This bug stems from the following logic in the debug window update procedure:
https://github.com/fantaisie-software/purebasic/blob/aa1dbc85256cdb3edb6348264b97fb73f12ccc11/PureBasicDebugger/DebugOutput.pb#L349-L351

When the text view's measured length equates to 0, `Range.NSRange\location` is assigned a value of -1.

Because `NSRange\location` is an `NSUInteger` (unsigned 32-bit or 64-bit integer depending on platform) value, macOS misinterprets -1 as the massive `2³²-1` or `2⁶⁴-1`. This freezes the process for an extended period of time.

## Solution

A simple fix would be to avoid taking action to scroll when the `Range.NSRange\location` equates to a negative value.

yuki-pb:
This PR resolves #224, avoiding an IDE crash on macOS when debugging empty strings.

### DIFF
--- a/PureBasicDebugger/DebugOutput.pb
+++ b/PureBasicDebugger/DebugOutput.pb
@@ -346,9 +346,16 @@ Procedure UpdateDebugOutputWindow(*Debugger.DebuggerData)
     CompilerIf #CompileMac
       ; Scroll the editor gadget to the bottom
       ; This function is very very slow, that's why we defer it (https://www.purebasic.fr/english/viewtopic.php?f=24&t=55924)
-      Range.NSRange\location = CocoaMessage(0, CocoaMessage(0, GadgetID(Gadget), "string"), "length") - 1
-      Range\length = 1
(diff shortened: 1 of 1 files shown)


fantaisie-software (approved):
Great, thanks for the fix !

PR #226, opened 2023-04-01
Some PB-IDE fixed for macOS

mk-soft-65:
**The macOS sizeMode function for NSToolBar no longer exists**. 
- This means that the icons (png) with the size 16x16 are scaled by the OS. To make the icons look better, the 16x16 icons are displayed center on icons 24x24.

**macOS ListIconGadget Column With**
- The ListIconGadget must be at least as wide as the first column. Otherwise the first column will not be displayed in width.

**Fix Text Color of Diff Files**
- In dark mode, the text is white and therefore not visible on marked items.

**Update Tools Design**
- Limit minimum height of buttons and string gadgets.
- Panel height increased because gadgets are not completely visible.
- ToolBars changed from TitleBar to Window content.


alphasnd (approved):
Nice, thank you !

PR #228, opened 2023-04-27
FormDesigner Updates

mk-soft-65:
Bugfix:
- The scrollbars of the workspace were sometimes not visible. The area is now always kept free for the scrollbars.
- Query of the scrollbar width is incorrect under Linux. Fixed value is now used.
- With multi selection of gadgets and multiple moving, the positions were not correct.
- The gadgets inside a scroll area gadget could be selected outside of the scroll area gadget.
- Mousewheel too slow. Value increased.

New:
- Multi selection of gadget with left pressed mouse and mouse movement. A selection frame is not displayed, but the selected gadgets are displayed immediately.


fantaisie-software (approved):
Looks good, thanks !

PR #229, opened 2023-04-28
FormDesigner Update 3

mk-soft-65:
Bugfix:
- IDE Crash at undo delete ScrollArea. Restore all gadget in right order

Linux:
- Added Filemanager 'nemo'. Its use from Mint Linux LMDE 5
- Added Filemanager 'pcmanfm' Its use from Raspberry


fantaisie-software (approved):
Looks good, thanks !

PR #230, opened 2023-05-07
Remove incorrect HTML tags `&nbsp;` in the help files

SicroAtGit:
Here, for example, in the array overview page:

pb-help-error


tajmone:
The named HTML entities for non-breaking space are correct, why are they not rendering as required?

Isn't there a way to save them? It might be preferable to replace them with literal NBSP characters, if nothing else works, so the intended spacing is preserved.

Naheulf:
I think there is two distinct issues with that.

1) As tajmone say the `&nbsp;` strings are valid html entities and so you should not see them. What is your OS and what which software do you use to display the help ? which help file format is it ?

2) That said entities looks like a failed attempt to align few random titles entries with the surrounding indented text. That why i think it's a good idea to remove them.
image

SicroAtGit:
@Naheulf

I have added detail information above.

fantaisie-software:
Looks good

PR #232, opened 2023-05-10
Improve array overview note text

SicroAtGit:
- "In PureBasic" removed from the English and German versions because it was not worded that way in the French version either.

- English sentence corrected.

- English and German version better adapted to the French version.


fantaisie-software (approved):
Perfect, thanks !

PR #236, opened 2023-07-30
Remove highlighting of inline ASM/C comments due to ambiguity with inline C

t-harter:
See https://www.purebasic.fr/english/viewtopic.php?t=81781


fantaisie-software (approved):
Perfect, thanks !

PR #237, opened 2023-07-30
Fix source parser for module prefix in 'Define' or similar keywords

t-harter:
See https://www.purebasic.fr/english/viewtopic.php?t=81506


alphasnd (approved):
Perfect, thanks !

PR #241, opened 2023-08-12
Remove an old Linux fix that causes problems when switching sources while loading a file

t-harter:
Removed a `FlushEvents()` call that messes up the file contents when switching tabs while loading many sources (for example on startup).

Note: The removed "fix" is very old and I cannot reproduce the behavior described in the old comments. So it probably does not apply anymore to the current version of Scintilla and other components. If the described behavior comes back we have to find a different solution for this.

See:
* https://www.purebasic.fr/english/viewtopic.php?t=76168
* https://www.purebasic.fr/english/viewtopic.php?f=18&t=71194
* https://www.purebasic.fr/english/viewtopic.php?f=23&t=69839

### DIFF
--- a/PureBasicIDE/SourceManagement.pb
+++ b/PureBasicIDE/SourceManagement.pb
@@ -130,6 +130,9 @@ Procedure UpdateSourceStatus(Modified)
   
 EndProcedure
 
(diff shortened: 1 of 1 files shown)


fantaisie-software (approved):
Perfect

PR #244, opened 2023-09-09
- Specialized CompareStringMemory() faster compare

fantaisie-software:
After fixing CompareMemoryString() to handle unicode properly, the routine is now slower, and it have a big impact on the IDE as we use it a lot. A specialized routine has been done to quickly handle ascii only case.


PureBasicIDE/Common.pb
```diff
@@ -2862,6 +2862,78 @@ CompilerIf #PB_Compiler_Debugger
   Global InDebuggerCallback = #False
 CompilerEndIf
 
+; Ascii only MemoryStringCompare to have much faster compare as we don't need unicode support for token parsing (https://www.purebasic.fr/english/viewtopic.php?t=78887)
+;
+Global  Dim NoCaseAsciiLookup(255)
```
Naheulf, line 2867: If that array only contain ASCII chars, shouldn't it be declared as `Global  Dim NoCaseAsciiLookup.c(255)`?

PR #245, opened 2023-09-14
- Changed the GetBucket() method to use a real hash and larger bucket…

fantaisie-software:
… as constants, structure etc can be very big list

### DIFF
--- a/PureBasicIDE/Common.pb
+++ b/PureBasicIDE/Common.pb
@@ -1797,7 +1797,7 @@ EndStructure
 ; special scope value (in addition to the debugger ones)
 #SCOPE_UNKNOWN = -1
 
-#PARSER_VTSize = 27   ; number of indexed entries per sorted array
+#PARSER_VTSize = 2048   ; number of indexed entries per sorted array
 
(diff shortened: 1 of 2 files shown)


t-harter:
The AutoComplete code assumes that each bucket represents words with a common start character. Changing it to a hash will break this. See for example here and other places in AutoComplete.pb: https://github.com/fantaisie-software/purebasic/blob/8faab0468c5682291241270057d7c11d930d58c5/PureBasicIDE/AutoComplete.pb#L385

I am not sure changing to a hash brings much benefits if the AutoComplete then has to scan the complete list again to find matching words.

Possible solutions:
* Keep the existing bucket mechanism but add a Map for direct access via the full name of an item too to speedup lookup of specific items.
* Replace the whole bucket mechanism by a prefix structure (like this https://www.purebasic.fr/english/viewtopic.php?p=586544). This could bring quite a benefit in all cases but would be a larger redesign.

fantaisie-software:
Thanks, I knew I had overlooked something :)

So a single map would be enough per file ? Could you try to draft it to see where it should be added ?

PR #246, opened 2023-09-15
- Changed the way the SCN_MODIFIED notification and ModifiedSinceUpda…

fantaisie-software:
…te flag is handled to reduce ScanLine() calls.

### DIFF
--- a/PureBasicIDE/HighlightingFunctions.pb
+++ b/PureBasicIDE/HighlightingFunctions.pb
@@ -1288,9 +1288,10 @@ EndProcedure
 
 Procedure.s GenerateQuickHelpText(Line$, Word$, Line, Column)
   
-  ; For the structure and prototype check we must ensure that the line data is up to date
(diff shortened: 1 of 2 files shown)


fantaisie-software:
It doesn't work as expected as case update isn't triggered if we change line an a modification occured (ie: endif -> EndIf). It should be easy to handle this special case. Another point is the SortParserData() routine which is getting very slow overtime as Parser_AddSorted is an O(n) routine which is getting very slow. One way to change this is to use a quicksort() routine to sort this and then add them at once in the bucket.

t-harter:
I have implemented a better way to store the sorted data as a tree in #250. This should solve the problem with `Parser_AddSorted()`. The `SortParserData()` still has to walk all source tokens once to process them in proper order to resolve context and scopes so it is still a bit dependent on overall size but it should be a lot better than the old solution. Can you try that together with your fix?

fantaisie-software:
RadixSort is enough for now, no need for this anymore

PR #248, opened 2023-09-20
Incomplete Preference Classic ToolBar

ChrisRfr:
See Topic: [Incomplete Preference Classic ToolBar](https://www.purebasic.fr/english/viewtopic.php?t=82483)
Since version 6.0, the Classic Toolbar in the preferences is incomplete, the theme icons are missing
And with the Default Toolbar there is an additional entry, Icon_24, Action_24 with indecipherable characters


fantaisie-software (approved):
Tested and works as exepected, thanks !

PR #249, opened 2023-09-24
Update Object_Gridbox.pb

Cyllceaux:
This is an easy fix for: https://www.purebasic.fr/english/viewtopic.php?t=82346

If you use "equal" in a GridBox, it will sometimes ignore the RequiredSize

### DIFF
--- a/DialogManager/Object_Gridbox.pb
+++ b/DialogManager/Object_Gridbox.pb
@@ -129,7 +129,9 @@ EndProcedure
 Procedure DlgGridBox_SizeRequest(*THIS.DlgGridBox, *Width.LONG, *Height.LONG)
   *Width\l  = 0
   *Height\l = 0
-  
+  MaxWidth = 0
+  MaxHeight = 0
+
   If *THIS\NbRows > 0
     
     For row = 0 To *THIS\NbRows-1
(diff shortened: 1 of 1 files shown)


fantaisie-software (changes requested):
Please check if MaxWidth/Height are correctly used

t-harter:
This is not the source code of the Dialog library, so it cannot be a bugfix for the linked bug report. Did you even test your fix?

DialogManager/Object_Gridbox.pb
```diff
@@ -146,15 +148,31 @@ Procedure DlgGridBox_SizeRequest(*THIS.DlgGridBox, *Width.LONG, *Height.LONG)
...
             *THIS\rowSize[row] = Max(*THIS\rowSize[row], Height)
+            MaxHeight = Max(*THIS\rowSize[row], MaxHeight)
           EndIf
         EndIf
         
       Next col
     Next row
+
+        If *THIS\colExpand = #Dlg_Expand_Equal
+                For row = 0 To *THIS\NbRows-1
```
fantaisie-software, line 164: The 'For row' is not necessary and should be removed

fantaisie-software (changes requested):

fantaisie-software:
Thanks ! The PureBasic lib has been patched the same way.

PR #250, opened 2023-09-26
Implement radix tree for faster source parsing and AutoComplete

t-harter:
Implementation of an indexed tree structure for parsed source tokens and AutoComplete for faster parsing and faster prefix lookup as discussed in #245.

This also removes the Preferences options to fill the AutoComplete list with all possible entries or all entries that match the first character. These options do not really make sense in the new model and probably had bad performance before anyway. So now the list is filled only with matches for the current typed word.

This is a larger change in the AutoComplete code so it needs some testing before being merged.


mk-soft-65:
Hello Timo,

please remove the AutoComplete.pb function AutoComplete_Redraw (for linux only). This sporadically leads to a recursive call of AutoComplete_WordUpdate.
See https://github.com/fantaisie-software/purebasic/pull/255

PR #251, opened 2023-10-02
Implement switch for accessibility features

t-harter:
Added new preferences switch to enable accessibility features. (it is under Preferences -> Editor -> Editing)

For now this only disables StatusBar ownerdraw (no bold text) to make it better to access via screen reader. Further accessibility features can be added depending on this switch in the future.

See https://www.purebasic.fr/english/viewtopic.php?t=82155


t-harter:
Some changes:
* Moved the setting to "Preferences -> General" to be easier to find.
* Enabling the setting also switches the menu icons off and selects the "Accessibility" color scheme
* Added auto-detection for Screen Readers on the first start

fantaisie-software:
Just tried and seems to work great. Thanks !

PR #253, opened 2023-10-07
fix(ToolsPanel): respect DPI scaling for panel hit-test region

### LINKED ISSUE #252: ToolsPanel with auto-hide does not respect DPI settings when hit-testing
## Platform

**PB:** 6.03 LTS beta 10 (145f8e74652a) (**Note:** I believe any DPI-enabled IDE version suffers the problem.)
**OS:** Windows 11, x64

## What's the problem?

When the ToolsPanel is in auto-hide mode, and the system is configured with non-100% DPI scaling, the region you must mouse-over in order to show the panel is incorrect.

Because the hit-testing region compares physical coordinates against logical coordinates, the ToolsPanel can be considered hovered-over far before the mouse is near it.

This can disrupt editing as the panel appearing can overlap code a user is trying to select. Also, when returning to the hidden state, the editor flickers, which can be jarring (especially given dark themes).

Below is a video demonstrating the problem:

https://github.com/fantaisie-software/purebasic/assets/127664937/0133ae9b-82ba-4673-b77a-2ad3da44aefc

## What is the expected behaviour?

The IDE should only show an auto-hidden ToolsPanel when the mouse actually hovers over it.

Below is a video demonstrating expected behaviour:

https://github.com/fantaisie-software/purebasic/assets/127664937/c7d5c076-5ec0-4a44-95d6-c303345ad390

## Why does this happen?

This bug stems from the following logic in the ToolsPanel auto-hide hit-test procedure:
https://github.com/fantaisie-software/purebasic/blob/145f8e74652a0323bc153f5cf60d63595959245c/PureBasicIDE/ToolsPanel.pb#L412-L426

The `MouseX` value returned from `WindowMouseX(#WINDOW_Main)` is measured in physical pixels, while both `EditorWindowWidth` and `ToolsPanelWidth` are measured in logical pixels. By mixing physical and logical pixels in comparisons and arithmetic, we see unexpected results.

## Solution

A simple solution would involve converting the physical X-coordinate to logical coordinate respecting the applied DPI scaling before comparing against other logical coordinates.

The [`DesktopUnscaledX(...)`](https://www.purebasic.com/documentation/desktop/desktopunscaledx.html) function should suffice to handle the needed conversion.

yuki-pb:
This PR resolves #252, ensuring DPI scaling is accounted for when hit-testing against the ToolsPanel in auto-hide mode.

### DIFF
--- a/PureBasicIDE/ToolsPanel.pb
+++ b/PureBasicIDE/ToolsPanel.pb
@@ -413,6 +413,12 @@ Procedure ToolsPanel_CheckAutoHide()
     
     If MouseX <> -1 And WindowMouseY(#WINDOW_Main) <> -1 ; do nothing if mouse is outside of the window
       
+      ; Convert cursor X position from physical to logical pixels to ensure that scale is
+      ; consistent with editor and tool-panel widths.
+      ; This must be done only after it's determined to be in-frame; unscaling prior will
+      ; result in failure at high scale factors (where -1 becomes 0).
+      MouseX = DesktopUnscaledX(MouseX)
+      
       If ToolsPanelSide = 0 ; right side
         Offset = EditorWindowWidth - MouseX
         ToolsPanelWidth = GadgetWidth(#GADGET_ToolsSplitter) - GetGadgetState(#GADGET_ToolsSplitter)


fantaisie-software (approved):
Perfect, thank you Yuki !

PR #254, opened 2023-10-11
fix(ToolsPanel): avoid flicker when hiding panel

yuki-pb:
This PR resolves a flickering issue mentioned in the conversation at https://github.com/fantaisie-software/purebasic/issues/252#issuecomment-1751737019, where the editing area and toolbar tends to flicker when hiding the ToolsPanel.

I tried a few approaches to avoid the flickering here, but temporarily pausing main-window painting with WM_SETREDRAW seemed to be the solution with least flickering. (Pausing just the editor area works, but then the toolbar flickers, so would have to be separately paused as well)

I have to admit I'm unfamiliar with the `SplitterGadget`, so there might be a better way to deal with the reparenting here without relying on WM_SETREDRAW.

### DIFF
--- a/PureBasicIDE/ToolsPanel.pb
+++ b/PureBasicIDE/ToolsPanel.pb
@@ -467,6 +467,11 @@ Procedure ToolsPanel_Hide()
       State = GetGadgetState(#GADGET_LogSplitter) ; somehow the reparenting makes the slider jump in the second splitter on linux, so only reset it after ResizeWindow()
     EndIf
     
+    CompilerIf #CompileWindows
+      ; Temporarily disable painting to avoid flickering while reparenting.
+      SendMessage_(WindowID(#WINDOW_Main), #WM_SETREDRAW, #False, 0)
+    CompilerEndIf
+    
     If ToolsPanelSide = 0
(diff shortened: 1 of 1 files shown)


ChrisRfr:
I haven't tried it yet, but I will try this weekend if I can find a bit of time.

Usually when you use WM_SETREDRAW Off/On, the [RedrawWindow](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-redrawwindow) function is then generally used with the flags to be adapted :

```
CompilerIf #CompileWindows
  ; Temporarily disable painting to avoid flickering while reparenting.
  SendMessage_(WindowID(#WINDOW_Main), #WM_SETREDRAW, #False, #Null)
CompilerEndIf
....
HideGadget()              
ResizeMainWindow()
....
CompilerIf #CompileWindows
  ; Restore painting to have updated visuals reflected.
  SendMessage_(WindowID(#WINDOW_Main), #WM_SETREDRAW, #True, #Null)
  RedrawWindow_(WindowID(#WINDOW_Main), #Null, #Null, #RDW_INVALIDATE | #RDW_ERASE | #RDW_ALLCHILDREN | #RDW_UPDATENOW)
CompilerEndIf
```

PR #255, opened 2023-10-26
IDE Update

mk-soft-65:
IDE-Crash:
```
[Debugger Error]  SelectElement(): SelectElement(): index value can not be negative.
[Debugger Error]  File: /home/pi/Documents/PB603/purebasic/PureBasicIDE/AutoComplete.pb (Line: 1621)
```

Fix: AutoComplete.pb:
- Remove AutoComplete_Redraw
- Caused a recursive call of AutoComplete_WordUpdate sporadically (Probably and not detectable, but no crash of the IDE anymore).

Fix LinuxMisc:
- Added for console lxterminal at Raspberry

Fix EditHistory:
- [Done] Gtk-CRITICAL **: 21:15:01.893: gtk_progress_bar_pulse: assertion 'GTK_IS_PROGRESS_BAR (pbar)' failed
- * Closing the window does not end the trigger for the ProgressBarGadget with the state #PB_ProgressBar_Unknown.

- [Done] Flickering of History Window
- * Calling SizeUpdate within Gadget Event Splitter repeatedly triggers a splitter event. A SizeUpdate may only be executed if the position changes.


PureBasicIDE/Common.pb
```diff
@@ -2865,4 +2865,22 @@ CompilerIf #PB_Compiler_Debugger
 CompilerEndIf
 
 UseMD5Fingerprint()
-UseCRC32Fingerprint()
\ No newline at end of file
+UseCRC32Fingerprint()
+
+; Limit ResizeGadget
+
+Procedure ResizeGadgetLimit(Gadget, x, y, Width, Height)
```
t-harter, line 2872: I think such hacks do not really belong in the IDE. This should be handled in the ResizeGadget() command directly if it does not work properly otherwise.

The IDE should not contain fixes for bugs in PB commands.

If it is to stay here then I think it should go into the `Macro.pb` file instead.

PureBasicIDE/EditHistory.pb
```diff
@@ -378,6 +377,7 @@ Procedure History_FlushEvents()
       
     Wend
     
+    SetGadgetState(#GADGET_HistoryShutdown_Progress, 0)
```
t-harter, line 380: This too feels more like a bug in the ProgressbarGadget command and should be fixed there and not in the IDE.

t-harter (commented):
I think you should split the unrelated issues into separate PRs. It is a bit too much a mixed bag this way.

mk-soft-65:
I still can't really get to grips with GitHub. I'll have to see how I can get a new PR with GitHub Desktop (macOS). My English isn't exactly the best.

PureBasicIDE/Common.pb
mk-soft-65: I wanted to remove the hack again.
Not even as a macro. Better customise it where ResizeGadget is called.

mk-soft-65:
I have removed the unimportant macro ResizeGadget.
The others should be taken over in the current IDE in the short term, as the IDE crashes very quickly under Linux (especially under Raspberry).

PR #258, opened 2024-01-16
SpriteBlendingMode AutoComplete context sensitive mode constants

ChrisRfr:
Some constants are missing for SpriteBlendingMode AutoComplete context sensitive mode


alphasnd:
Thanks !

PR #259, opened 2024-01-17
Use internal mkdir command on Windows

ChrisRfr:
On Linux and OS X, use "mkdir" (mkdir.exe) instead of mkdir, to make it works, as it is currently.
On Windows, use internal mkdir command, mkdir.exe does not exist and internal mkdir does not support -p switch

Otherwise, on Windows we get this error:
`"mkdir.exe" -p "D:\PureBasic_Opensource\PureBasic_Windows_beta2_X64_6.10/Themes" - 
'"mkdir.exe"' n’est pas reconnu en tant que commande interne ou externe, un programme exécutable ou un fichier de commandes.`

Not verified on Linux and OS X that it continues to work as expected!

### DIFF
--- a/PureBasicIDE/CompilerInterface.pb
+++ b/PureBasicIDE/CompilerInterface.pb
@@ -77,7 +77,7 @@ CompilerIf #CompileWindows
     #COMPILER_UNICODE    = " /UNICODE" ; still needed to set in older compilers
   CompilerEndIf
   
-  #COMPILER_FileFormat = #PB_UTF8
+  #COMPILER_FileFormat = #PB_Ascii
 CompilerElse
   CompilerIf #CompileMac And Not #SpiderBasic
     #COMPILER_STANDBY    = " --standby -f -ibp" ; extra flags for osx only
@@ -95,7 +95,7 @@ CompilerElse
     #COMPILER_UNICODE    = " --unicode" ; still needed to set in older compilers
   CompilerEndIf
   
-  #COMPILER_FileFormat = #PB_UTF8
+  #COMPILER_FileFormat = #PB_Ascii
 CompilerEndIf
 
 
--- a/PureBasicIDE/Makefile
+++ b/PureBasicIDE/Makefile
@@ -188,13 +188,22 @@ Build/DefaultTheme.zip: data/DefaultTheme/*
 	rm -f Build/DefaultTheme.zip
 	zip -jq Build/DefaultTheme.zip data/DefaultTheme/*
 
-# Use "mkdir" instead on mkdir, so on Windows it search mkdir.exe on the PATH instead of using the internal CMD mkdir.
-# It works on Linux and OS X as well.
+# On Linux and OS X, use "mkdir" (mkdir.exe) instead of mkdir, to make it works
+# On Windows, use internal mkdir command, mkdir.exe does not exist and internal mkdir does not support -p switch
+# %OS% is the type of windows: Windows_NT
 #
+ifeq ($(OS),Windows_NT)
 $(THEMEFOLDER)/SilkTheme.zip: data/SilkTheme/*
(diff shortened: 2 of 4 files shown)


ChrisRfr:
I'm still having trouble understanding GitHub
My 2nd commits was automatically added to this Pull Request.
I would have liked 2 separate Pull Requests as they are unrelated.
I should have isolated them in their own branch, but I didn't know it.
You can delete this PR if it doesn't suit the needs.

PureBasicIDE/CompilerInterface.pb
```diff
@@ -77,7 +77,7 @@ CompilerIf #CompileWindows
     #COMPILER_UNICODE    = " /UNICODE" ; still needed to set in older compilers
   CompilerEndIf
   
-  #COMPILER_FileFormat = #PB_UTF8
+  #COMPILER_FileFormat = #PB_Ascii
```
fantaisie-software, line 80: This shouldn't be changed. The Windows compiler is now unicode and communicate using UTF-8

PureBasicIDE/CompilerInterface.pb
```diff
@@ -95,7 +95,7 @@ CompilerElse
     #COMPILER_UNICODE    = " --unicode" ; still needed to set in older compilers
   CompilerEndIf
   
-  #COMPILER_FileFormat = #PB_UTF8
+  #COMPILER_FileFormat = #PB_Ascii
```
fantaisie-software, line 98: Same, linux/osx already use UTF-8

PureBasicIDE/Makefile
```diff
@@ -188,14 +188,23 @@ Build/DefaultTheme.zip: data/DefaultTheme/*
...
 
+else
+$(THEMEFOLDER)/SilkTheme.zip: data/SilkTheme/*
+	"mkdir.exe" -p "$(THEMEFOLDER)"
+	rm -f "$(THEMEFOLDER)/SilkTheme.zip"
+	zip -jq "$(THEMEFOLDER)/SilkTheme.zip" data/SilkTheme/*
+endif
+
 $(COLORTABLE): data/ColorTable.xml
 	cp -f data/ColorTable.xml "$(COLORTABLE)"
```
fantaisie-software, line 210: Never change the Makefile, it's used by the build system. May be I should put a comment on top of it. You should use the MakeWindow.cmd on Windows

fantaisie-software (changes requested):

PureBasicIDE/Makefile
ChrisRfr: Oh sorry Fred, the goal was to help, not to waste your precious time.
It's probably best to delete this Pull Request and to leave it up to you to solve the 3 issues
Can you delete it please, I'm afraid I do worse than better on my Github's friend.

Thanks for the comments even if here:
On windows changing #COMPILER_FileFormat #PB_UTF8 to #PB_Ascii, looks good here to accept special chars (É). With UTF-8 the Linker ends in error if there are any special chars.
There are probably still a few tools (maybe vc/link.exe) that communicate using Ascii.

I understand for the makefile, I'll use the quick & easy method now: MakeWindows.cmd <YourPureBasicPath> .
But then, it's good to know that the official method doesn't fully work on Windows. "mkdir" -p is not recognized (mkdir.exe does not exist on windows) and so the Themes folder is not (re)created. At the same time, it's already there.

Finally, the only thing that seems ~ right is PeekAscii(*Debugger\CommandData) => PeekUTF8(*Debugger\CommandData) in IDEDebugger and StandaloneDebugger.pb
For special characters in error or warning messages in the debugger.

fantaisie-software: Don't worry, no problem here. About mkdir, you need to put the unixtools path in your PATH to have the unix command instead of the Windows one.

I will patch the PeekUTF8 stuff, so I will close this request

PR #267, opened 2024-03-24
FormDesigner Added WebViewGadget

mk-soft-65:
Added WebViewGadget to FormDesigner
- Common variable / constant name and counter for WebGadget and WebViewGadget. So compatible for old form files.
- Change control name WebView to Web for WebGadget
- Added control name WebView for WebViewGadget
- Added Flags for WebGadget
- Added Flags for WebViewGadget
- Added EventType for ListIconGadget and ExplorerListGadget


fantaisie-software (approved):
Thanks !

PR #271, opened 2024-06-23
Linux Help Window Fix

mk-soft-65:
PB-Help crashed on Ubuntu 24.04
Move update help contents to main event loop.


fantaisie-software:
Thanks !

PR #273, opened 2024-06-28
Documentation grammar correction for 'Bound'

spikey-pb:
The past tense of 'bind' is 'bound' not 'binded'.


alphasnd (approved):
Thanks ! I also changed it in the .desc file for the IDE quick help

PR #274, opened 2024-06-29
QT Subsystem support

t-harter:
Adds support for creating a QT version of the IDE. There are no big TODOs left right now but it still needs some testing. This is why this is a draft for now. Feedback is welcome.

To build with QT, just set the environment variable `PB_QT=1` before compilation (make sure `PB_GTK` is not set as well). This will produce an executable called `purebasic_qt`.

Notes:
* Unfortunately it was not possible to do this in pure PB, so the `QtHelpers.cpp` is needed for some platform specifics. This means you need GCC and the QT devel packages to build this.
* The preferences files between the QT and GTK versions are not compatible (because all shortcut constant values differ). So either delete your preferences file or use a separate one with the `-p` option when starting the QT IDE.


alphasnd:
Cool stuff ! May be you can change the name for the pref file when QT is active so it's easier to test

alphasnd:
I successfully compiled and run the IDE on Ubuntu 22.04, it looks solid so far, very good work ! I noticed a crash on the Color chooser, last panel ('name') and using the scrollbar. I got a few refresh issue but it's probably due to VirtualBox, so I will test on a real box later. Will do some more tests and let you know, a very promising start !!

alphasnd:
I will list the issues I found here:

- crash on the Color Chooser (tab name) when scrolling the name
- when using a Tree in the 'Explorer' tool, double cliking on a file do nothing
- the 'Help' window is properly displaying content, but there is no index and the << and >> button doesn't work
- Debugger -> Show Error log doesn't work (the error log is still visible)

fantaisie-software:
Is it possible to rename the shortcut keys in the .prefs with '_qt' ? So we can use the same pref file for all the backends and I could merge it to have a live test for 6.20

PR #279, opened 2024-07-28
Enable unhide of tools pane on mac os

Shardik-PB:
The procedure panel doesn't open automatically when the cursor is at the right edge of the IDE window. It also doesn't open when clicking onto "<<" on the vertical bar at the right edge. When deactivating "Automatically hide the Panel", the procedure panel opens at once.

This bug occurs in PB 6.10 and 6.11 in at least these MacOS versions:
- MacOS 10.15.7 'Catalina'
- MacOS 13.6.7 'Ventura'
- MacOS 14.5 'Sonoma'

Automatical unhiding works like a charm in PB 6.04 and previous versions.

In Linux Mint 21.3 and Windows 10 this bug doesn't occur at all.

Bug reported in PureBasic forum "Bugs IDE":
https://www.purebasic.fr/english/viewtopic.php?p=623251#p623251


Shardik-PB:
I finally succeeded to upload the change file for the pull request. During many hours of trying I always received the error message
`remote: Support for password authentication was removed on August 13, 2021.`
This error message is totally misleading. My authentication token expired and after generating a new token, I was finally able to successfully push the pull request from my MacOS console with git push...

tajmone:
@Shardik-PB:

> This error message is totally misleading. My authentication token expired and after generating a new token, I was finally able to successfully push the pull request from my MacOS console with git push...

Indeed, the authentication tokens can be confusing. I remember when GitHub introduced them, I was faced by the same situation as you and had to spend a few hours Googling to understand what was going on. Furthermore, some repositories require different types of tokens, e.g. with more privileges to access GH Actions, etc., and in such cases you won't be able to contribute if you don't have a token matching the required features (regardless of whether you're using them or not).

fantaisie-software:
Thanks for the fix !

PR #282, opened 2024-10-01
Linux build target executables now lowercase "purebasic"

kenmo-pb:
The PureBasicIDE project file has pre-defined build targets for Windows, Mac, and Linux. See #102 

Now that I'm using PB Linux, I realize the official packaged executables are all lowercase. The IDE is  `purebasic`, not `PureBasic`. Because Linux paths are case sensitive, I've corrected the two Linux build targets.


fantaisie-software:
Perfect !

PR #283, opened 2024-10-01
Linux: bring Help window to foreground after creating

kenmo-pb:
On Linux*, it took two presses of `F1` to see the Help window. The first press created the Help window, but it always appears behind the IDE, subsequent presses brought it to the foreground. This Linux-only change brings `#WINDOW_Help` to the foreground immediately after creation.

* ElementaryOS 7.1 (Ubuntu 22.04 based)


fantaisie-software:
Looks good, thanks !

PR #284, opened 2024-10-01
Linux: handle LeftTab event to allow tabbing backward through ProjectInfo

kenmo-pb:
On Linux*, with a project open in the IDE, I can `Ctrl+Tab` forward through all open tabs with no problem, but `Ctrl+Shift+Tab` backward gets "stuck" on the ProjectInfo tab.

It seems `Ctrl+Shift+Tab` throws a GDK key value `$FE20` (`LeftTab`) plus `Ctrl` and `Shift` modifiers. This patch handles those events, similar to how `ScintillaShortcutHandler()` handles `$FF09`, and posts the appropriate `#MENU` response.

⚠️ _Can others with more PB Linux and GDK experience confirm this bug, and review this bugfix? Thanks!_

* ElementaryOS 7.1 (Ubuntu 22.04 based)

### DIFF
--- a/PureBasicIDE/ProjectManagement.pb
+++ b/PureBasicIDE/ProjectManagement.pb
@@ -626,6 +626,20 @@ CompilerIf #CompileLinux
     Debug "Key event: " + *Event\keyval
     If *Event\keyval = #GDK_Return
       PostEvent(#PB_Event_Gadget, #WINDOW_Main, Gadget, #PB_EventType_LeftDoubleClick)
+      
+    ElseIf *Event\keyval = $FE20 ; #GDK_LeftTab
+      If (*Event\state & (1 << 2)) ; Ctrl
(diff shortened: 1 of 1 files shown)


fantaisie-software:
Is it an IDE only issue or should the PureBasic libs needs to be patched instead ?

kenmo-pb:
@fantaisie-software I'm not sure if it's an issue of gadget library or menu library or IDE code... There must be a reason this is being handled as GDK key events rather than regular PB shortcuts. Maybe @t-harter can check this out? All I know is the `Shift` seems to change the GDK Tab event to a GDK LeftTab event, at least on my distro.

fantaisie-software:
Juts tested and it works as expected. I don't think it's a lib issue, as the project info tab set the focus elsewhere it's handled differently

PR #285, opened 2024-10-06
Integration of the 'Multicolored Procedure List'

ghost:
Integration of the 'Multicolored Procedure List' similar to the RSBasic tool.

Forum-Thread:
https://www.purebasic.fr/english/viewtopic.php?t=85421

Functions:

- Hide the module name.
- Highlight the entry in the list to match the cursor position in the editor.
- Scroll the entry in the list to the visible area.
- Prevent the automatic unfolding of folded code.
- Color the font of the selected entry.
- Color the background of the selected entry.
- Remove the colors of the selected entry.
- Copy the entries to the clipboard. Only the procedures, all entries, with arguments.
- Filter the entries.

Procedures, markers and macros can be colored. Issues cannot be colored directly.
The text of the colored entry is compared with the beginning of the other entries and matching entries are colored.
Issues can be colored with a marker ";- TODO", for example. The marker can then be deleted.

Activation of the list:
  Preferences -> ToolsPanel -> Procedure Browser -> Configuration: Multicolored Procedure List

Added resources:
- The texts for the ToolTips have been added for all languages.
-  The catalog files for Russian are also available.
- The icons for the buttons have been added to the Default- and SilkTheme.
- The settings of the list are saved in the preference file 'PureBasisc.prefs':

```
[ProcedureBrowser]
Sort = 0
Prototype = 0
; New
Multicolor = 0
HideModuleName = 0
HighlightProcedure = 0
HighlightTimer = 1000
ScrollProcedure = 0
EnableFolding = 0
SwitchButtons = 0
; 
; New
[ProcedureBrowser_ItemColor]
; Nicht verwendeter Schlüßel, die Werte sind: Schriftfarbe, Hintergrundfarbe, Text.
1 = -1,8454143,-- DeclareModule --
2 = -1,8454143,-- Module --
3 = -1,8454016,-- Event-Procedure.
...
```

Modified source files:
-  Common.pb
-  Language.pb
-  ProcedureBrowser.pb
-  ThemeManagement.pb
-  UserInterface.pb

Tested with:
-  Windows 7 32-/64-Bit
-  Debian 12 64-Bit Cinnamon
-  Raspberry Pi OS


fantaisie-software (changes requested):
It's very good work. Could you enable it by default ? It basically superseed the old browser, with more functionalities. Also the filter should be 'realtime' IMHO (each time the stringgadget content change, the result is updated),   so we shouldn't need the 'X' button to clear the field

ghost:
I have made the changes, but have not yet updated them here.

The multicolored procedure list is displayed as default if the entry does not yet exist in the preferences file.
The filter is updated as you type. The Clear button has been removed.
In a standalone procedure list window, the focus is not set back to the editor when one of the new buttons is clicked. This is otherwise irritating with overlapping windows.

The Ctrl-X/V/C keys do not work in the filter, the entries are inserted in the editor.
I don't know yet if I can program this correctly.

Peter

ghost:
I have updated the changes on git.
The function for Ctrl-X/C/V in the filter is still missing.

Peter

fantaisie-software:
Looks good to me ! About the Ctrl+X/C/V it a general issue in the IDE which need to be tackled (we should remove the shortcurts when scintilla loose the focus and adding them back when the focus is back)

PR #286, opened 2024-10-07
Debugging IDE Drag-and-drop

ChrisRfr:
When debugging the IDE with a new tab then drag and drop a project, it crashes with: 
[ERROR] WindowEvent() can Not be called from a 'binded' event callback.


fantaisie-software (approved):
Looks good, thanks !

PR #287, opened 2024-10-09
Debugger max Structures Fields to Display in Tooltip

ChrisRfr:
Calculate the maximum number of lines when debugging structures to always display the tooltip up or down with the maximum number of fields possible.

Currently, with MaxLine = 25 hardcoded, no possibility to display more lines for large structures if there is enough space.
If there isn't enough space at the bottom or top, the tolltip is Not displayed at All. 
Large structure tooltip is often not displayed when the cursor is on a line in the middle of the editor, no problem when the cursor is at the top or bottom.

With the calculation, the number of lines to be displayed is adjusted (with a 1 line margin to be on the safe side) to the available up or down space. 
And so make sure the tooltip is always displayed with the maximum number of structure fields displayed.
 
Tested on Windows only but not on Linux, Mac


fantaisie-software:
I tested on Linux and OSX and it seems to work as expected, thanks !

PR #288, opened 2024-10-09
Remove /NOEXT switch from non-Windows build targets

kenmo-pb:
The `/NOEXT` commandline switch should not have been added to the Mac and Linux build targets, there it interprets `/‍NOEXT` as a file path and attempts to load it (and fails, oops).


fantaisie-software:
Well spot ! Thanks

PR #290, opened 2024-11-20
Better Linux darkmode TabBar

ErlendRovik:
First commit, a more natural darkmode tabbar for PureBasic IDE.

Just a couple of alterations or colors and such in two files..


fantaisie-software:
Tested and it indeed look much better ! Thanks for the contribution

PR #291, opened 2024-11-23
Linux/Mac: Normalize keyboard shortcuts to uppercase values

kenmo-pb:
Problem: (Linux/Mac, not Windows) The **ShortcutGadget** used to assign keyboard shortcuts can return uppercase or lowercase values depending on whether `Shift` or `CapsLock` was active. For example `Alt+'A'` or `Alt+'a'`. Only the uppercase matches a `#PB_Shortcut_*` constant. This causes (at least) two problems:

1. When converting value to string (see shortcut table in IDE preferences), lowercase-valued shortcuts will be missing the letter. For example `Alt+'a'` will show as just `Alt+`.
2. Since you can generate two different shortcut values (try it with `CapsLock` on or off), you can assign the same key combination twice in the IDE table, without warning. Then it seems only the uppercase-value shortcut with work!

Ideal solution: It would probably be best to solve this in the ShortcutGadget library. Could it always return a consistent value regardless of `Shift` and `CapsLock` state?

**Simple IDE fix for now**: This patch adds `NormalizeShortcut()` which remaps lowercase values to uppercase. It can be expanded to more cases. Currently, it only has an effect on Linux/Mac, and it only has an effect on alphabet letter shortcuts.

Forum thread about this: https://www.purebasic.fr/english/viewtopic.php?t=85404

### DIFF
--- a/PureBasicIDE/Preferences.pb
+++ b/PureBasicIDE/Preferences.pb
@@ -471,6 +471,7 @@ Procedure LoadPreferences()
     Read.s MenuItem$
     Read.l DefaultShortcut
     KeyboardShortcuts(i) = ReadPreferenceLong(MenuItem$, DefaultShortcut)
+    KeyboardShortcuts(i) = NormalizeShortcut(KeyboardShortcuts(i))
   Next i
   
   ; a little fix to not have a shortcut twice, since i switched two shortcut values in the defaults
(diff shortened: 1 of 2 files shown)


fantaisie-software:
I will probably fix it before 6.20 in lib for cross plateform consistency

kenmo-pb:
OK, thanks. :+1: I saw you were preparing for 6.20 and I thought I'd submit this workaround quickly. I have a small new feature to submit too, maybe for 6.20 or later.

Just be warned, I think this ShortcutGadget issue affects more than just alphabet characters, for example key `1` returns `#PB_Shortcut_1` but `Shift+1` returns `#PB_Shortcut_Shift | '!'` (US keyboard) which would be more complicated to remap...

fantaisie-software:
I fixed the lib for the next release, so we can close this topic

PR #297, opened 2025-01-07
Russian-Catalog-Encoding

ChrisRfr:
Message from AZJIO who seems not to have access
The Russian compiler.catalog, debugger.catalog and libraries.catalog files must be in utf-8 without BOM. Now with BOM and because of this the text is in English.

My fault, when I converted, I probably only looked at editor.catalog to find out the encoding for the other languages, UTF8+Bom for this one.


fantaisie-software (approved):
Thanks

PR #298, opened 2025-01-07
Added #PB_Editor_TabNavigation to the constants data for EditorGadget().

trypsynth:
The new #PB_Editor_TabNavigation constant added in 6.20 wasn't added to the ConstantsData.pb file for EditorGadget(), leading to it not showing up in autocomplete.


fantaisie-software (approved):
Thanks

PR #300, opened 2025-01-19
Update List.txt

chikega:
"an own", is ungrammatical in English because "own" is a possessive adjective that requires a determiner like "my," "your," "his," "her," "its," "our," "their", or "one's" (e.g., "my own", "your own").

I believe the intended meaning of the sentence is that SelectElement() allows you to move to a specific position in a linked list without manually iterating through the list using a loop.

### DIFF
--- a/Documentation/English/List.txt
+++ b/Documentation/English/List.txt
@@ -773,7 +773,7 @@
   
 @Description
   Change the current list element to the element at the specified position.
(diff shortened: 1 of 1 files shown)


fantaisie-software:
Thanks for your commit, but the doc is handled on another repo, so we won't accept anymore doc changes here (as it's too hard to merge)

PR #301, opened 2025-01-25
Fix spelling mistakes in the English help

SicroAtGit:
Tool used and its parameters:
`codespell -ew Documentation/English/`

What I also noticed is that the following files are not translated, but are kept in German, and they do not exist in the French directory either:
- https://github.com/fantaisie-software/purebasic/blob/devel/Documentation/English/MainGuide/regform2.txt
- https://github.com/fantaisie-software/purebasic/blob/devel/Documentation/SpiderBasic/English/MainGuide/regform2.txt


fantaisie-software:
Thanks for your commit, but the doc is handled on another repo, so we won't accept anymore doc changes here (as it's too hard to merge)

fantaisie-software:
Thanks !

SicroAtGit:
@fantaisie-software

Now you have accepted the PR after you removed your comment in which you mentioned that no more doc changes would be accepted on GitHub because there is a different repo (SVN?) for them (glad I didn't waste my time).

Have you now changed your decision again and doc changes are still allowed on GitHub? If doc changes are no longer allowed on GitHub, this should be clearly stated in the [`CONTRIBUTING.md`](https://github.com/fantaisie-software/purebasic/blob/devel/CONTRIBUTING.md) file, or better the doc files should be removed then, so that no one would think of touching the doc files.

Can you also say what should happen to the mentioned `regform2.txt` files? Should they continue to be in German in other languages as well?

fantaisie-software:
Yes I changed my mind because it wasn't respectful from the work already done, so I will continue to merge the private SVN into public Git which is a pain to do.

About regform2.txt I don't think it should be in german in other languages

PR #307, opened 2025-02-19
Add new options to Form Designer

CalDymos:
I have made a small change to 'codeviewer.pb' and 'opensave.pb' so that a custom 'parent' for the form can be passed using the suffix '='.
Now the following is possible

```Wnd = OpenWindow(#PB_Any, x, y, width, height, "", #PB_Window_BorderLess, customHwnd)```

in which the following is entered in **parent**:  `=customHwnd`

Then `customHwnd` can be defined elsewhere in the code, e.g. as follows.

`customHwnd = GetWindow_(GadgetID(#ScrollArea),#GW_CHILD))`

I have also added the possibility to set custom flags for the forms.

With these two options, it is now possible to display a form in a ScrollAreaGadget, for example, without having to modify the form designer code every time.

### DIFF
--- a/PureBasicIDE/FormDesigner/codeviewer.pb
+++ b/PureBasicIDE/FormDesigner/codeviewer.pb
@@ -1010,10 +1010,10 @@ Procedure.s FD_SelectCode(contentonly = 0, testcode = 0)
             line + ", "
           EndIf
           
-          If Not FindString(FormWindows()\parent, "(")
+          If Asc(FormWindows()\parent) <> '=' 
             line + "WindowID(" + FormWindows()\parent + ")"
           Else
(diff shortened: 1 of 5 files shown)


CalDymos:
It should now work correctly and remains compatible with “Form Designer” code with and without this customization

CalDymos:
I have added the possibility to set the '#PB_Frame_Container' flag in the Form Designer for FrameGadgets and the FrameGadget can then be used as a container in the Form Designer.

fantaisie-software:
It looks good, I will test your changes and merge it if all is good. If you think you didn't have tested it enough, just tell me so I won't merge.

PR #308, opened 2025-02-19
Debug ToolTip width enlarged to available space

ChrisRfr:
The Debug ToolTip width enlarged according to the available space between the mouse position (to be aligned) and the right desktop border, rather than 100 hard-coded chars.


fantaisie-software (approved):
Looks good thanks !

PR #311, opened 2025-02-23
Replaced 'brackets' with 'parentheses' in english documentation.

trypsynth:
The correct word (where replaced) is parentheses, not brackets. :)


fantaisie-software (approved):
Nice !

PR #313, opened 2025-03-05
Fix for single-line "Remove Comments" when caret is at start of line

kenmo-pb:
Fix for "Remove Comments". It now does the expected uncomment behavior if the caret is at the very start of a line (column 1) with no text selection.

See https://www.purebasic.fr/english/viewtopic.php?t=84667


fantaisie-software (approved):
Looks good !

PR #314, opened 2025-03-05
Show file name instead of full path, in "Modified on Disk" and "Deleted on Disk" dialogs

kenmo-pb:
Sometimes a file's full path overflows the "Modified on Disk" dialogs.

This patch only shows the file's name, not the full path, in the `ModifiedOnDisk1` and `ModifiedOnDisk2` and `DeletedOnDisk` dialogs. In 99% of cases, this is probably enough info. In fact the `Modified` dialog when closing already does this.

Other suggestions in the forum are compacting the file path to the available width, or dynamically growing the window size.

See https://www.purebasic.fr/english/viewtopic.php?t=60349


fantaisie-software (approved):
Looks good, thanks !

PR #319, opened 2025-04-12
Add ElementaryOS terminal support to LinuxMisc

kenmo-pb:
This adds `io.elementary.terminal` to the searched terminals in `LinuxMisc.pb`, so the internal debugger can work in Console build mode.

See https://www.purebasic.fr/english/viewtopic.php?t=86701


fantaisie-software (approved):
Looks good, thanks !

PR #321, opened 2025-07-05
Fix macOS IDE crash on exit and Fix FormDesigner Custom Flags

mk-soft-65:
From the logic, the error should not occur. Guess it's because of the external scintilla library callback when the parent window is closed.


fantaisie-software:
Looks good, thanks for the fixes !

fantaisie-software (approved):

PR #325, opened 2025-07-24
Localisation File Consistency Checker

spikey-pb:
* Add a new tool 'LFCheck' for checking localisation files to the 'tools' subfolder.

### DIFF
--- a/PureBasicIDE/tools/LFCheck.pb
+++ b/PureBasicIDE/tools/LFCheck.pb
@@ -1,0 +1,427 @@
+﻿; --------------------------------------------------------------------------------------------
+;  Copyright (c) Fantaisie Software. All rights reserved.
+;  Dual licensed under the GPL and Fantaisie Software licenses.
+;  See LICENSE and LICENSE-FANTAISIE in the project root for license information.
+; --------------------------------------------------------------------------------------------
+
+; Localisation File Consistency Checker.
+
+; Usage:
+; LFCheck /u /r="reference file path" /t="target file path" /l="log file path"
+; /u - Check for untranslated text.
+; /r= - Reference catalog path, omit this to reference the IDE's default language table.
+; /t= - Target catalog path.
+; /l= - Log problems to a file.
+
+; Please note:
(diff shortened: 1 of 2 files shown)


PureBasicIDE/tools/LFCheck.pb
```diff
@@ -0,0 +1,427 @@
...
+
+;- Enumerations
+EnumerationBinary Origin
+  #Reference
+  #Target
+EndEnumeration
+#Both = #Reference | #Target
+
+EnumerationBinary Options
+  #OptReferenceFile
```
t-harter: The usual naming for constants in the IDE code is more `#OPTION_ReferenceFile` (enum type in caps separated with `_` from the actual name.

PureBasicIDE/tools/LFCheck.pb
```diff
@@ -0,0 +1,427 @@
...
+Structure ASTRING
+  State.I
+  RefCount.I
+  TargetCount.I
+  Reference.S
+  Target.S
+EndStructure
+
+;- Variables
+Global.S ReferenceDesc = #DefTable$, ReferenceFile, TargetFile, LogFile
```
t-harter: Type names are not all caps in the IDE code. Also, we consistently use `Variable$` for string variables (you have a bit of a mix here) and leave out the `.i` outside of structures since it is the default type.

The rule is this:
* Strings always as `Variable$`
* Integers always as `Variable`
* Floats or Doubles always as `Variable.d`. Not just at the declaration but also when used to make it more visible that it is not an integer variable. Floating point variables are rare, though.

t-harter (commented):
I added some notes on code style. It is not that important here since this is a separate tool, but it still might make sense to adjust it to fit the rest of the code better.

I also suggest to rename the tool to `LanguageCheck.pb` or `LanguageFileCheck.pb` if you prefer. The name "LFCheck" really does not say much if you don't know what the file is about.

PureBasicIDE/tools/LFCheck.pb
spikey-pb: Updated.

PR #326, opened 2025-07-26
Form Designer Missing Localisations

spikey-pb:
* Update some popup menus to use language table entries.


t-harter:
You have included the `LFCheck.pb` both in this PR and in #325. This will cause a conflict when merging.

spikey-pb:
I'm not sure where I went wrong there.  I think I've corrected it now though.

PR #328, opened 2025-08-12
New Logos

spikey-pb:
- Updated logos for the splash screen and about dialog in the new website style.
- Changed splash screen version text to black for PureBasic.


fantaisie-software:
I'm not sure about this change TBH :)

PR #329, opened 2025-08-21
(Preferences) If a shortcut is already used, ask if user wants to reassign it, instead of rejecting the change

kenmo-pb:
Just a small change to make the Shortcut Preferences more user-friendly.

If the user tries to assign a shortcut which is already used, instead of rejecting it with a "Shortcut is already used..." dialog, change that to a "Shortcut is already used... Reassign it?" Yes/No requester, allowing them to accept it immediately.

This saves the user some clicks and scrolling and keypresses... You don't need to find the conflicting shortcut in the list, unassign it, return to original shortcut, assign that again.

### DIFF
--- a/PureBasicIDE/Language.pb
+++ b/PureBasicIDE/Language.pb
@@ -1667,6 +1667,7 @@ DataSection
   Data$ "ProceduresUpdate", "Trigger Update of Procedure & Variable Viewer"
   
   Data$ "AllreadyUsed",     "The shortcut you specified is already used by" ; DO NOT FIX TYPO: AllreadyUsed
+  Data$ "ReassignPrompt",   "Should the shortcut be reassigned?"
   Data$ "ExternalTool",     "External Tool"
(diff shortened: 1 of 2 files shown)


PureBasicIDE/Language.pb
```diff
@@ -1658,6 +1658,7 @@ DataSection
   Data$ "ProceduresUpdate", "Trigger Update of Procedure & Variable Viewer"
   
   Data$ "AllreadyUsed",     "The shortcut you specified is already used by" ; DO NOT FIX TYPO: AllreadyUsed
+  Data$ "ReassignPrompt",   "Reassign shortcut?"
```
t-harter: I would prefer "Should the shortcut be reassigned?" instead.

kenmo-pb:
String changed, rebased to latest `devel`, pushed again :+1:

t-harter:
Thanks.

PR #332, opened 2025-09-02
- Set caret location after a template insert using the ^ symbol.

spikey-pb:


PureBasicIDE/ScintillaHighlighting.pb
```diff
@@ -3501,17 +3499,31 @@ CompilerIf #CompileWindows | #CompileLinux | #CompileMac
...
+    ; Embed insertion into a single transaction, otherwise the insertion mark will reappear in an Undo action.
+    SendEditorMessage(#SCI_BEGINUNDOACTION, 0, 0)
+    
+    CurrentPos = SendEditorMessage(#SCI_GETCURRENTPOS, 0, 0)
     Converted$ = Space(StringByteLength(String$, Format))
     PokeS(@Converted$, String$, -1, Format)
     SendEditorMessage(#SCI_REPLACESEL, 0, @Converted$)
+    
+    ; Remove marker and move caret, if a position is specified.
+    MarkerOffset = FindString(String$, "^")
```
t-harter: You need to take care of possible UTF-8 encoding here (see the Format variable). The scintilla commands usually take byte offsets, not character positions. Best is to do some testing with accented or other special characters and try both UTF8 and Ascii to make sure both work right.

t-harter (commented):
Nice addition.

Could you also add a note to document this in the English help for the IDE?
See here: https://github.com/fantaisie-software/purebasic/blob/devel/Documentation/English/Reference/ide_tools.txt

PureBasicIDE/ScintillaHighlighting.pb
t-harter: Also, I just noticed that the procedure you modified (`InsertCodeString()`) is used in multiple places in the IDE. So placing this caret move code in there could have some side effect on another use case, maybe.

Why not add a new parameter to the procedure to enable/disable this behavior? This way it will not affect anything outside of the template functionality. You can make it an optional parameter so existing code does not need to be changed.

spikey-pb: OK, done.

fantaisie-software (approved):
Thanks !

PR #333, opened 2025-09-06
Form Designer Version Checking

spikey-pb:
- Warn if an opened pbf file doesn't contain the form designer title line.

- Warn about a version upgrade and compatibility issues when opening a file from an older version.

- Warn about a downgrade when opening a file from a newer version.

- Added a new section to the preference dialog to control these warnings.


fantaisie-software (approved):
Looks good to me, nice addition

PR #335, opened 2025-09-16
Form Designer Missing Flags

spikey-pb:
- Add some missing flags to FD reference lists, notably for OpenGLGadget.

- Add some missing events to FD reference lists.


fantaisie-software (approved):
Perfect, thanks !

PR #342, opened 2026-01-22
Sort arrays, lists, and maps in variableViewer!

ChrisRfr:
In variableViewer, sorts arrays, lists, and maps by their names by default.
Unless there is a reason that I am not seeing


fantaisie-software (approved):
Looks good

PR #346, opened 2026-01-29
Less resize on splitters changes

ChrisRfr:
UpdateSourceContainer() on splitters changes only if there is a real change in the splitters positions.  For fewer processes, resizing, redraw.

However, wouldn't it be better if, in the Gadget Library, only Splitter events with a real state change (position) were processed?
Moving the mouse Horizontally over a Horizontal Splitter does nothing and is useless, it shouldn't send an event (resize > redraw for us). And vice versa for the Vertical Splitter

### DIFF
--- a/PureBasicIDE/UserInterface.pb
+++ b/PureBasicIDE/UserInterface.pb
@@ -2069,11 +2069,10 @@ Procedure MainWindowEvents(EventID)
     EventGadgetID = EventGadget()
     Select EventGadgetID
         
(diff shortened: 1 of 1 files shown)


fantaisie-software (changes requested):
Looks good to me, could you remove the Debug statements ?

I don't think we will patch the splitter gadget as introducing a cache could lead to other issues (I tried and it failed in a few cases)

ChrisRfr:
I've removed the debugging statements that were only there to make testing easier.
Thank you for your explanation on the Splitter events.

fantaisie-software (approved):

PR #347, opened 2026-02-02
Feature/form editor explicit gadget

CalDymos:
## Summary

This PR adds support for **explicit gadget IDs** & **explicit window IDs** in the PureBasic Form Designer property field, using the syntax:

- `ID13310=13310` or `Gadget1=1008` (variable=explicit numeric ID)

The generated code will keep the **variable name** as-is, while the **enumeration constant** gets the explicit assignment:

```purebasic
Enumeration FormGadget
  #ID13310=13310
EndEnumeration
```

## Motivation

Some projects require stable, predefined numeric IDs (e.g. compatibility with existing code, external bindings etc.). Until now, the Form Designer only supported automatic IDs (#PB_Any) or non-explicit enumeration constants, making it hard to keep IDs deterministic across edits/saves.

## Behavior / Rules

The property field `variable` may contain `Name=Number` to define an explicit mapping.
Code generation outputs `#Name=Number` in Enumeration FormGadget / FormWindow.
Gadget/window creation uses `#Name` or `Name` for Procedure Open...
**Backward compatible / non-breaking:** If a property value contains no `=`, the Form Designer behaves **exactly as before** (same parsing, same code generation).

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,24 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Run MakeWindows.cmd",
+            "type": "shell",
+            "command": "powershell",
+            "args": [
+                "-ExecutionPolicy", "Bypass",
+                "-NoProfile",
+                "-Command",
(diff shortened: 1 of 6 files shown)


fantaisie-software (approved):
Looks good to me

PR #348, opened 2026-02-03
Docs/form designer enhancements

CalDymos:
This PR updates the PureBasic documentation to cover recent Form Designer enhancements:

- Explicit gadget/window IDs via the Variable property using Name=Number syntax (PR #347).
- Raw parent expressions by prefixing the parent value with '=' and the Custom Flags window property (PR #307).
- FrameGadget acting as a container via #PB_Frame_Container (PR #307).

The changes are documentation-only and do not affect IDE behavior.

Related:
- PR #307
- PR #347


fantaisie-software (changes requested):
Thanks for the PR ! You should remove the binary .chm which shouldn't be commited

CalDymos:
I removed the .chm file and updated .gitignore to ignore all .chm files.

fantaisie-software:
Thanks !

PR #349, opened 2026-02-13
UG_PB_Any2 Tutorial Corrections

spikey-pb:
- Corrected an incorrect pointer type in DestroyTrackWindow.

- Adjusted the layout of the Date Window so the Calendar isn't clipped on Windows 11.

- Fixed a minor typography problem.


fantaisie-software (approved):
Thanks !

PR #350, opened 2026-02-21
Linux: Detect elementaryOS file manager "Files"

kenmo-pb:
Similar to #319 

Ubuntu-based elementaryOS uses a non-standard name for its file manager. This adds `io.elementary.files` to the searched names, so that IDE functions like "Show in Folder" will work on this OS.


fantaisie-software (approved):
Looks good

PR #352, opened 2026-02-25
Fix FontRequester() missing args so Style is passed in correctly

kenmo-pb:
Calls to `FontRequester()` from Preferences window were missing the **Flags** and **Color** args, so **Style** was not being passed in correctly, and always defaulted to Regular style. Fixed!


fantaisie-software (approved):
Nice fix !

PR #356, opened 2026-03-22
Prevents the tab from moving when activated

dreisiebner:
PB v6.40b4, macOS Tahoe 26.3.2

When switching between tabs with a mouse click, a tab may sometimes move even after the mouse button has been released.
This happens when the tabs are sometimes displaying the ‘Error Log’ and sometimes not. In such cases, the ‘LeftButtonUp’ event is not triggered, or rather, it does not reach the TabBarGadget.

See the forum post:
IDE - Switching to a different code tab moves the tab
https://www.purebasic.fr/english/viewtopic.php?t=88543

Solution by STARGATE:
https://www.purebasic.fr/english/viewtopic.php?p=653197#p653197


fantaisie-software (approved):
Thanks for the fix !

PR #341, opened 2026-01-16
Sort and Filter Project Files

ChrisRfr:
In Project Management and Project Panel tool
Add a checkbox to sort the project files on Filename and Folder\Filename and a StringGadget to filter the displayed items.

PR #362, opened 2026-03-26
Resize #GADGET_ProjectInfo_FrameFiles following PR #341

ChrisRfr:
Tiny update to resize #GADGET_ProjectInfo_FrameFiles height on ProjectManagement.pb, following PR [#341](https://github.com/fantaisie-software/purebasic/pull/341) "sort and Filter Project Files". 
The frame height wasn't adjusted to fit tits content with the new #GADGET_ProjectInfo_FilterInput,  my fault!


kenmo-pb:
You fixed it before I even got to GitHub to report it :+1: Thanks for this new filter feature by the way.

Minor thing, but any idea why Ctrl+A to Select All doesn't work in that filter StringGadget? At least on Linux. I think the Ctrl+A keyboard shortcut is intercepted somewhere else in the IDE code and I wonder if it could be passed along so it behaves like any other typical StringGadget.

ChrisRfr:
> Minor thing, but any idea why Ctrl+A to Select All doesn't work in that filter StringGadget? At least on Linux. I think the Ctrl+A keyboard shortcut is intercepted somewhere else in the IDE code and I wonder if it could be passed along so it behaves like any other typical StringGadget.

It's not really essential for filters, but it doesn't cost much to remove the four main shortcuts (Ctrl+A, Ctrl+C, Ctrl+X, Ctrl+V) on Focus and recreate them on LostFocus.
It's already done here for UserInterface, I need to add that as well for ProjectPanel and ProcedureBrowser and I'll create a pull request, up to Fred to merged it or not

PR #368, opened 2026-04-02
Add/Remove Scintilla shortcuts for MacOS too

ChrisRfr:
Add/Remove Scintilla shortcuts on #SCN_FOCUSIN, #SCN_FOCUSOUT for MacOS too, using #PB_Shortcut_Command (rather than #PB_Shortcut_Control for windows and Linux).

It should be good, but I don't have a Mac to test it!

### DIFF
--- a/PureBasicIDE/Commandline.pb
+++ b/PureBasicIDE/Commandline.pb
@@ -415,21 +415,22 @@ Procedure ParseCommandline()
           DontCreateExtensions = 1
           
         Case "/LOCAL"
+          ; Keep PreferencesFile$, TemplatesFile$,... if already defined using -p “PreferenceFile”, -t “TemplatesFile”,... before /PORTABLE switch
           Directory$ = GetPathPart(ProgramFilename())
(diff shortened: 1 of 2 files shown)


kenmo-pb:
Even simpler, you can just use `#PB_Shortcut_Command` for all, because on Windows and Linux it is equivalent to `#PB_Shortcut_Control`. You can check the values yourself. See the help for `AddKeyboardShortcut()` about this :+1:

ChrisRfr:
I agree again (simpler but probably less descriptive)👍
But I do it like it is in ShortcutManagement.pb (line 423+) with the CompilerIf #CompileLinux | #CompileWindows -> CompilerElse for MacOS
Use #PB_Shortcut_Command could be good, by adding the comment: "; #PB_Shortcut_Command will act like #PB_Shortcut_Control on non-macOS "
Done, thanks

ps; Thanks [kenmo-pb](https://github.com/kenmo-pb) for following the filter added to the project management and to help improve it with the default behavior for strings, now 👍

fantaisie-software:
Looks good !

PR #371, opened 2026-04-02
Cmdline preserve previous settings with /PORTABLE switch at the end

ChrisRfr:
Keep PreferencesFile$, TemplatesFile$,... if already defined using -p “PreferenceFile”, -t “TemplatesFile”,... before /PORTABLE switch

However, with the -P, -T, -A, and -H switches before /PORTABLE, it is no longer truly portable, in the literal sense of the word.

I'll let you decide whether it's right for you or not

### DIFF
--- a/PureBasicIDE/Commandline.pb
+++ b/PureBasicIDE/Commandline.pb
@@ -415,21 +415,22 @@ Procedure ParseCommandline()
           DontCreateExtensions = 1
           
         Case "/LOCAL"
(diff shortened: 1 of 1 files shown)


fantaisie-software:
Looks better this way IMHO, as the commandline shouldn't depends of args orders
-p "test.perfs" /PORTABLE
should be the same than
/PORTABLE -p "test.perfs" 

Thanks !

fantaisie-software (approved):

PR #376, opened 2026-04-17
Add/remove keyboard shortcuts for FilterInput StringGadget

ChrisRfr:
Add/remove the 4 main keyboard shortcuts (Ctrl+A, Ctrl+C, Ctrl+X, Ctrl+V) to restore the standard behavior for FilterInput StringGadget when they gain/lose focus

To replace the generic way, used previously, when the Scintilla component loses/gains focus
But with this generic way, these shortcuts (cut, copy, paste) no longer work in Form Designer, see: https://www.purebasic.fr/english/viewtopic.php?t=88746


fantaisie-software:
No, there generic way is much better as all the tools are affected. We just need to add them when the form is activated and remove it when it loose focus

ChrisRfr:
I understand of course. And you're right not to give up on the generic way :)

But I felt responsible and  6.40 is out now. I wanted to suggest a fix as soon as possible and
I don't know how to do this in FormDesigner, or I just need some help figuring out where it should be added.

Of course, you can close this PR

ChrisRfr:
I reworked this PR to match what you wanted using the generic way and added the creation of shortcuts when Form Designer is activated and removed them when it loses focus.

Rather than writing the code, to add or remove shortcuts multiple times, I added two new procedures, AddStringShortcuts() and DelStringShortcuts(), to ShortcutManagement.pb.

ScintillaHighlighting.pb becomes: #SCN_FOCUSIN > AddStringShortcuts() and #SCN_FOCUSOUT > DelStringShortcuts()

In addition, for the FormDesigner:
AddStringShortcuts() is called when FormDesigner opens, in SourceManagement.pb procedure ChangeActiveSourcecode() and also in mainevents.pb procedure FD_SetDesignView()
DelStringShortcuts() is called when switching tabs from FormDesigner to ProjectInfo. For the other tabs (scintilla), it has already been done with #SCN_FOCUSIN event. The same applies for the Menu option: "switch Code/Design View"

Note: In Form Designer, design view, shortcuts are Not added/removed when switching tools (in the right-hand panel) between "Form" to "Procedures" or "Project". This is intentional, it is not essential here and it helps avoid bloating the source code.

PR #390, opened 2026-09-07
New tool bar icons

spikey-pb:
New Toolbar Icons

Adds entries to the themed icons table for:
- 'Show in Folder'
- 'Paste as Comment'

Adds new Silk Theme icons for:
- 'Clear Markers'
- 'Syntax Check'
- 'Find Previous'
- 'Show in Folder'
- 'Paste as Comment'


tutankino:
the real problem It's not the icons themselves, but their size—they're too tiny. You can't even see them on a 4K screen. There should be an option to increase their size, including for the fonts in the various menus, which are almost unreadable on a 4K screen.

spikey-pb:
I am working on a 24px icon set when I have time but that's 163 icons not 5!  I haven't got a 4K monitor to test with but I'm far from convinced that even that will make a significant difference on a 4K monitor and I think that's a limit at the library level at present (as far as I can tell).  To go larger might require a custom toolbar implementation which is a totally different project altogether.

PR #396, opened 2026-09-13
Fix FormDesigner macOS

mk-soft-65:
Fix limit gadgets to bottom


fantaisie-software:
If it really fix it, you can remove all the compilerif block

fantaisie-software:
Is there an associated bug thead on the forum ?

mk-soft-65:
> Is there an associated bug thead on the forum ?

Yes, on Bugs IDE
No Link. Forum at time very slow ;(

